PR #4, opened 2017-10-24
pkp/pkp-lib#2102 Implement SelectAll feature for SelectListPanel.

NateWr:
- Adds SelectListPanelSelectAll component.
- Implements select all for submission list panels.
- Refactors ListPanel notices into ListPanelNotice component.

### DIFF
--- a/src/components/ListPanel/submissions/SubmissionsListPanel.vue
+++ b/src/components/ListPanel/submissions/SubmissionsListPanel.vue
@@ -23,6 +23,12 @@
 			/>
 		</div>
 		<div class="pkpListPanel__body -pkpClearfix pkpListPanel__body--submissions">
+			<select-list-panel-select-all
+				v-if="showSelectAll"
+				:label="i18n.selectAllLabel"
+				:checked="selectAllChecked"
+				@toggle="toggleSelectAll"
+			/>
 			<submissions-list-filter
 				v-if="currentUserCanFilter"
 				@filterList="updateFilter"
--- a/src/components/ListPanel/submissions/SubmissionsListPanel.vue
+++ b/src/components/ListPanel/submissions/SubmissionsListPanel.vue
@@ -23,12 +23,6 @@
 			/>
 		</div>
 		<div class="pkpListPanel__body -pkpClearfix pkpListPanel__body--submissions">
-			<select-list-panel-select-all
-				v-if="showSelectAll"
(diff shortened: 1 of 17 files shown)


src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -23,6 +23,12 @@
 			/>
 		</div>
 		<div class="pkpListPanel__body -pkpClearfix pkpListPanel__body--submissions">
+			<select-list-panel-select-all
+				v-if="showSelectAll"
+				:label="i18n.selectAllLabel"
+				:checked="selectAllChecked"
+				@toggle="toggleSelectAll"
+			/>
```
bozana: why this select-all feature in the SubmissionsListPanel?

src/components/SelectListPanel/SelectListPanelItem.vue
```diff
@@ -67,26 +62,23 @@ export default {
...
+	methods: {
 		/**
-		 * Set an item to selected when loaded
+		 * Emit an event to select or de-select this item
 		 */
-		if (this.item._initializeSelected) {
-			this.selected = this.inputType === 'radio' ? this.inputValue : true;
-		}
+		toggle: function () {
+			this.$emit('toggle', this.inputValue);
```
bozana: where is then this selected/de-selected?

src/components/ListPanel/submissions/SubmissionsListPanel.vue
NateWr: Ahh! You're right. I don't know what I was thinking. :)

https://github.com/NateWr/ui-library/commit/e8e13b4b83b2cd61b0009666764e2469592dbfa1

src/components/SelectListPanel/SelectListPanelItem.vue
NateWr: The parent component, `SelectListPanel`, listens for the event and toggles the value here:

https://github.com/NateWr/ui-library/blob/select_all/src/components/SelectListPanel/SelectListPanel.vue#L27

Vue.js follows a pattern where data needs to be changed at the highest-level component that controls it. Data that is "passed down" to a child component shouldn't be modified in that child component. When a child component wants to modify that data, it needs to emit an event to the parent, which updates the data.

https://vuejs.org/v2/guide/components.html#Composing-Components

PR #14, opened 2018-03-20
pkp/pkp-lib#3130 Improve access control for dual-assignments

NateWr:


src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -100,14 +100,14 @@ export default {
...
-			return pkp.userHasRole(['manager', 'subeditor', 'assistant']);
+			return pkp.userHasRole(['ROLE_ID_MANAGER', 'ROLE_ID_SUB_EDITOR', 'ROLE_ID_ASSISTANT']);
 		},
 
 		/**
 		 * Does the current user have a role which can create a new submission?
 		 */
 		currentUserCanAddSubmission: function () {
-			return pkp.userHasRole(['manager', 'subeditor', 'assistant', 'author']);
+			return pkp.userHasRole(['ROLE_ID_MANAGER', 'ROLE_ID_SUB_EDITOR', 'ROLE_ID_ASSISTANT', 'ROLE_ID_AUTHOR']);
```
bozana: This seems not to work, so maybe it should be: pkp.const.ROLE_ID_MANAGER, pkp.const.ROLE_ID_SUB_EDITOR, pkp.const.ROLE_ID_ASSISTANT, pkp.const.ROLE_ID_AUTHOR

NateWr: Good catch!

src/components/ListPanel/submissions/SubmissionsListItem.vue
```diff
@@ -205,7 +215,7 @@ export default {
 		 * @return bool
 		 */
 		currentUserCanViewInfoCenter: function () {
-			return pkp.userHasRole(['manager', 'subeditor', 'assistant']);
+			return this.userAssignedRole([pkp.const.ROLE_ID_SITE_ADMIN, pkp.const.ROLE_ID_MANAGER, pkp.const.ROLE_ID_SUB_EDITOR, pkp.const.ROLE_ID_ASSISTANT]);
```
bozana: Hmmm... I think we removed the assistant role, or?

asmecher: No, that's the generic term for copyeditor/layout editor/etc.

bozana: Hmmm... @asmecher, what do you think? 
I mean the assistant role is not allowed any more to access the submission history i.e. function viewHistory in controllers/informationCenter/InformationCenterHandler.inc.php, s. contructor there now... Or?

asmecher: :) sorry, that was an obvious answer. @NateWr? OTOH the assistant can now view the file information centre but not the submission one, correct?

NateWr: Hmm, good catch. So, in my tests, it looks like an assistant can view the Notes in the information center, but not the History. They can view them from the submissions list.

However, in the submission workflow, an assistant who clicks on Editorial History gets a role error, and can't even access the Notes.

Is there a reason the assistants shouldn't have access to the Editorial History? If they shouldn't have access, then I'll remove access to the Notes and the link in the submission list. If they should, I'll restore them to the op handler so they can get in from the workflow.

Any thoughts?

asmecher: The assistant role has quite a swath of potential users -- copyeditors, layout editors, potentially translators, external agencies, etc. So they'll definitely need access to the metadata (which used to leave in the information centre, and I'm pretty sure is the reason they were allowed in at all), but I don't think they'll need access to the history etc. And I think of the notes area as a private scratch pad for privileged users (editors/managers and arguably assigned section editors). So I'd lean towards excluding them from the information center altogether.

NateWr: Makes sense. I've implemented that in https://github.com/NateWr/pkp-lib/commit/4f6f094f617eeda2b4765aabf9880d748f69ff33 and https://github.com/NateWr/ui-library/commit/c0aed50447201bcde61dbb91dd7f39d4e206ce17

PRs have been rebased and tests are running. I'll merge if they pass.

PR #25, opened 2018-11-13
Revert formatting

NateWr:
@asmecher A heads up on this revert in case it happens again. It looks funny, but the mix of rendered HTML and code-example HTML here is correct. It displays something like this:

selection_034


asmecher:
Gotcha, thanks!

PR #31, opened 2019-05-16
pkp/pkp-lib#2508 Re-add reviewer cancel feature

asmecher:


src/components/ListPanel/users/SelectReviewerListItem.vue
```diff
@@ -145,6 +145,13 @@
 					</template>
 					{{ i18n.declinedReviews }}
 				</list-item>
+				<list-item>
+					<template slot="value">
+						<icon icon="times-circle-o" :inline="true" />
```
NateWr: Splitting hairs here, but let's try to differentiate the icons used for cancelled and declined. Replace `times-circle-o` here with `ban`.

PR #33, opened 2019-08-20
pkp/pkp-lib#4844 Editorial report

jonasraoni:
Here I've just made a small modification in the code which loads the data.

### DIFF
--- a/src/components/Chart/DoughnutChart.vue
+++ b/src/components/Chart/DoughnutChart.vue
@@ -1,0 +1,60 @@
+<script>
+import { Doughnut, mixins } from 'vue-chartjs';
+
+export default {
+	extends: Doughnut,
+	mixins: [mixins.reactiveProp],
+	props: {
+		chartData: {
+			type: Object,
+			required: true,
+		},
+		chartOptions: {
+			type: Object,
+			default: function () { return {}; },
+		},
+	},
+	computed: {
+		/**
+		 * Merge default options with options passed as a prop
+		 *
+		 * @return Object
+		 */
+		options: function () {
+			return {
+				responsive: true,
+				maintainAspectRatio: false,
+				cutoutPercentage: 80,
+				layout: {
+					padding: {
+						top: 32,
+						bottom: 32,
+						left: 32,
+						right: 32,
+					},
+				},
+				legend: {
+					display: false,
+				},
+				elements: {
+					arc: {
+						borderWidth: 0,
+					},
+				},
+				...this.chartOptions,
+			};
+		},
+	},
+	mounted: function () {
+		this.renderChart(this.chartData, this.options);
+	},
+};
+</script>
+
+<style lang="less">
+@import '../../styles/_import';
+
+.chartjs-render-monitor {
+	background: @bg-anchor;
+}
+</style>
--- a/src/components/Chart/DoughnutChart.vue
+++ b/src/components/Chart/DoughnutChart.vue
@@ -53,8 +53,4 @@ export default {
 
 <style lang="less">
 @import '../../styles/_import';
-
-.chartjs-render-monitor {
-	background: @bg-anchor;
-}
 </style>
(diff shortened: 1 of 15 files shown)


src/components/Statistics/Statistics.vue
```diff
@@ -234,9 +234,12 @@ export default {
 						return;
 					}
 
-					self.timeSegments = r.timeSegments;
-					self.items = r.items;
-					self.itemsMax = r.itemsMax;
+					// Reassign any existing data property
+					for (const key in r) {
+						if (key in self.$data) {
```
jonasraoni: Just needed the component to reload some other data, so I changed the handler to be more generic and added a small check to limit what can be updated.

src/components/Chart/DoughnutChart.vue
```diff
@@ -0,0 +1,60 @@
...
+		this.renderChart(this.chartData, this.options);
+	},
+};
+</script>
+
+<style lang="less">
+@import '../../styles/_import';
+
+.chartjs-render-monitor {
+	background: @bg-anchor;
```
NateWr: This should be unnecessary. The component itself should use a transparent background so it can be reused anywhere.

src/components/Statistics/Statistics.vue
NateWr: Makes sense. However, the UI component is now tightly coupled to the API. The API response must contain property names that match the component's property names. It's not always easy to do, but we try to avoid this as much as possible.

In this case, I think it makes sense to create new components. `Statistics.vue` will become a _base_ component and we can create two components, `StatisticsSubmissions.vue` and `StatisticsEditorial.vue`, which extend it.

This will have the side-benefit of splitting our dependencies, so that `DoughnartChart` and `LineChart` (as well as the styles) are only loaded for the components that need them. It doesn't make a difference to our bundle size now, because we're not splitting the bundle per page. But eventually we'll do that so having separate components will help there. And keeping the styles for each page separate will also be useful.

What I would suggest is changing the success callback function to a `method` in `Statistics.vue`:

```
$.ajax({
  ...
  success: function (r) {

    // Only process latest request response
    if (self._latestGetRequest !== this._uuid) {
      return;
    }

    this.handleApiResponse(r);
  },
  ...
});
```

In `Statistics.vue` this can be a dummy method:

```
methods: {
  ...
  handleApiResponse(response) {
    return;
  }
}
```

(Or if you want to get fancy you can throw and catch an error message that tells the developer they need to extend the component.)

Then `StatisticsEditorial.vue` can be a lean component that just implements what's necessary:

```
<script>
import Statistics from './Statistics.vue';
import DoughnutChart from '@/components/Chart/DoughnutChart.vue';

export default {
  extends: Statistics,
	components: {
		DoughnutChart
	},
	data: function () {
    return {
      // any additional stats
    };
  },
  methods: {
    handleApiResponse(response) {
      // do the assignment here
    }
  }
}
</script>

<style lang="less">
@import '../../styles/_import';

// Specific styles here
</style>
```

And the same for `StatisticsEditorial.vue`. Does that seem like an ok approach?

src/docs/examples/Statistics/ViewEditorialStatistics.vue
```diff
@@ -0,0 +1,169 @@
...
+					}
+					return true;
+				});
+
+				this.isLoading = false;
+			}, 1000);
+		},
+	},
+};
+</script>
```
NateWr: Thanks for documenting this! It's a lot of work but worth it! Often, this documented version falls out of sync with what's in the Smarty template, so every now and then it's good to copy that template over and rebuild it to make sure it's in sync with how it's actually used.

But even when out of sync with the exact template used in OJS/OMP, it's still a useful guide to how the component is meant to be used. :+1:

src/components/Statistics/Statistics.vue
jonasraoni: That's fine! I'm just not sure if this component is generic enough to be in this repository 🤔

NateWr: Yeah, so we are putting all of our Vue.js components into this repository. We don't just use it for base components. We also use it to document every component, so either way it should go into this library.

NateWr:
Replaced by https://github.com/pkp/ui-library/pull/51

PR #34, opened 2019-08-23
pkp/pkp-lib#4168 show dateStatusModified in submission list

forgive38:


forgive38:
I understand your point of vue.
I did it because, 2 of our journals ask for this feature, and especially, because I would like to have no differences between our version and ojs-stable.

NateWr:
> I would like to have no differences between our version and ojs-stable.

I understand. I'll work to try to get this back in before too long. I'm just overloaded with the [versioning](https://github.com/pkp/pkp-lib/projects/15) work right now. Feel free to remind me.

PR #37, opened 2019-09-25
pkp/pkp-lib#4168 Add last activity date to submissions list

NateWr:


src/components/ListPanel/submissions/SubmissionsListItem.vue
```diff
@@ -720,6 +723,14 @@ export default {
...
 			this.mask = null;
 		},
+
+		formatDate: function (str) {
+			if (!str) { return ''; }
+			str = new Date(str);
+			return str.getFullYear() + '-' + ((str.getMonth() < 9) ? '0' : '') + (str.getMonth() + 1) + '-' +
+				((str.getDate() < 10) ? '0' : '') + str.getDate();
+		},
+
```
NateWr: Apologies, I should have caught this before. The `dateStatusModified` will _always_ be returned in the format `YYYY-MM-DD HH:MM:SS`. So you should be able to just do a `substring`:

```
item.dateStatusModified.substring(0, 10)
```

forgive38: I propose this : 
```
let dateLocale = $.pkp.app.currentLocale.replace('_', '-');
return str.toLocaleDateString(dateLocale);
 ```
Format Date in current language

NateWr: :+1: perfect! Can you add a filter to the global mixins file (https://github.com/pkp/ui-library/blob/master/src/mixins/global.js), so that it's available to every component?

forgive38: good idea
I add localizeDate in global.js

NateWr: Looks good. I preferred your initial implementation as a `filter`. Do you mind making this a global `filter`? It is more explicit about its purpose. :+1:

forgive38: but with a filter, I'm not sure you can use it like this : 
{{ __('lastActivity', { dateStatusModified | localizeDate }) }}
or I have to modify __ function ?

NateWr: Hmm, ok. :( Let's go ahead with the method then.

PR #45, opened 2019-10-17
Support versioning in authorDashboard

ajnyga:
Pull request for
https://github.com/pkp/pkp-lib/issues/4877
https://github.com/pkp/pkp-lib/issues/3758

### DIFF
--- a/src/components/Container/WorkflowContainer.vue
+++ b/src/components/Container/WorkflowContainer.vue
@@ -21,7 +21,7 @@ export default {
 	data() {
 		return {
 			contributorsGridUrl: '',
-			disableSave: '',
+			canEditPublication: '',
 			csrfToken: '',
 			currentPublication: null,
 			editorialHistoryUrl: '',
@@ -369,7 +369,7 @@ export default {
 
 				// Add/remove save button depending on publication status or user permissions
 				form.pages = form.pages.map(page => {
-					if (publication.status === pkp.const.STATUS_PUBLISHED || this.disableSave) {
+					if (publication.status === pkp.const.STATUS_PUBLISHED || this.canEditPublication) {
 						delete page['submitButton'];
 					} else {
 						page.submitButton = {label: this.i18n.save};
--- a/src/components/Container/WorkflowContainer.vue
+++ b/src/components/Container/WorkflowContainer.vue
@@ -369,7 +369,7 @@ export default {
 
 				// Add/remove save button depending on publication status or user permissions
 				form.pages = form.pages.map(page => {
-					if (publication.status === pkp.const.STATUS_PUBLISHED || this.canEditPublication) {
+					if (publication.status === pkp.const.STATUS_PUBLISHED || !this.canEditPublication) {
 						delete page['submitButton'];
 					} else {
 						page.submitButton = {label: this.i18n.save};
--- a/src/components/Container/WorkflowContainer.vue
+++ b/src/components/Container/WorkflowContainer.vue
@@ -229,6 +229,7 @@ export default {
 				title: this.i18n.uploadFileModal,
 				url: this.uploadFileUrl,
 				closeCallback: () => {
+					pkp.eventBus.$emit('refreshRevisionsGrid');
 					this.$refs.uploadFileButton.$el.focus();
 				}
 			};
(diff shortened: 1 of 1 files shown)


src/components/Container/WorkflowContainer.vue
```diff
@@ -522,6 +537,12 @@ export default {
 			this.loadRepresentationsGrid(this.workingPublication);
 		}, 1000);
 
+		/**
+		 * Load forms
+		 */
+
+		 this.setPublicationForms(this.workingPublication);
```
NateWr: Ah ok, so yeah this makes sense. In theory, the forms should load in the appropriate state so that we don't have to do this after mounting -- for example, there may be a few `ms` when the form's save button appears, and this creates headaches for the tests.

But I'm happy to leave this as-is for now. The one thing I would suggest is to move this into the `created()` lifecycle method rather than `mounted()`. This will fire pre-mount so we shouldn't end up writing the DOM for the forms, then immediately changing and rewriting it. It should also avoid any `ms`s that might cause test funniness.

src/components/Container/WorkflowContainer.vue
```diff
@@ -21,6 +21,7 @@ export default {
 	data() {
 		return {
 			contributorsGridUrl: '',
+			disableSave: '',
```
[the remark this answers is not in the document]
ajnyga: check for the renaming

Also the authors grid needs to be refreshed then I think?

src/components/Container/WorkflowContainer.vue
ajnyga: I was just thinking whether https://github.com/pkp/pkp-lib/issues/5160 affected the way things are getting loaded at the moment.

src/components/Container/WorkflowContainer.vue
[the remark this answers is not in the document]
ajnyga: hmm, the refresh is actually tricky. Basically we would need to emit what kind of change has happened when saving the stage participant form.

So here https://github.com/pkp/pkp-lib/blob/master/controllers/grid/users/stageParticipant/StageParticipantGridHandler.inc.php#L352 you need to set $canEditPublication to true or false depending on the change that just happened, basically pass a new value there.

Then in https://github.com/pkp/pkp-lib/blob/master/js/controllers/grid/users/stageParticipant/StageParticipantGridHandler.js#L36 you need to pass that value onwards to the new globalEvent

And finally in https://github.com/pkp/ui-library/blob/master/src/components/Container/WorkflowContainer.vue#L482 you need to check if that globalEvent is fired and if it is, then refresh the $canEditPublication value and the contributors grid.

Is this even remotely how it should happen?

Question: how do you access the json events->data contents in the StageParticipantGridHandler.js constructor?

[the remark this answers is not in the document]
NateWr: On second thought, we don't need this. Because if the user has permission to edit a stage assignment, they have permission to edit the publication. There may be some situation where a section editor changes their own stage assignment to block editing, but this will be rare and when they go to make an edit they'll still get blocked, so I think the UI update is not important.

I was in the midst of describing how to fire a global event directly from the PHP response. So I'll include that just for future reference. Where you see the following:

```php
return DAO::getDataChangedEvent($userGroupId);
```

You can emit a global even here by doing:

```php
$json = DAO::getDataChangedEvent($userGroupId);
$json->setGlobalEvent('current-user-permissions-changed', ['canEditPublication' => false]);
return $json;
```

So then you should be able to listen in Vue.js for `pkp.eventBus.$on('current-user-permissions-changed', (args) => { ... });`.

src/components/Container/WorkflowContainer.vue
NateWr: Maybe, but if I recall correctly the publication forms always load with the save button, so this will probably need to be called on load now or the initial form data needs to be adjusted. But find to call his on `created()` for now.

NateWr: It looks like there are some whitespace issues here. Also, can you move this up to the `created()` lifecycle method, rather than `mounted()`? It will update the data earlier in the cycle and I think it may be slightly more performant because no shadow dom stuff will be running by then.

ajnyga: totaly forgot about this, fixed now!

PR #47, opened 2019-10-24
pkp/pkp-lib#2617 Added badge count to My Queue in submissions

thinkbulecount2:
From this issue: https://github.com/pkp/pkp-lib/issues/2617


src/components/Tabs/Tabs.vue
```diff
@@ -24,6 +24,9 @@
 				<template v-else>
 					{{ tab.label }}
 				</template>
+				<template v-if="tab.badge">
+					<badge :label="tab.label">{{ tab.badge }}</badge>
```
NateWr: In this case we can get rid of the `:label` property because it repeats what's already above. So a screen reader will read out "my queue my queue 12".

Often badges are used without any other text, and in such cases a `label` is important. But in this case it's best to leave it out.

src/docs/components/Tabs/previews/PreviewTab.vue
```diff
@@ -1,9 +1,9 @@
 <template>
 	<tabs label="Default demo of tabs component">
-		<tab label="First tab" id="first">
+		<tab label="First tab" id="first" badge="12">
 			First tab content
 		</tab>
-		<tab label="Second tab" id="second">
+		<tab label="Second tab" id="second" badge="2">
```
NateWr: This looks great. Now that you've got the `<badge>` working, can you split this preview out from the default preview?

If you look at the code structure in the `docs/components/Tabs/`, you'll see that there's `PreviewTab.vue`, `PreviewSide.vue`, etc. So in this case we want `PreviewTab` to be the most basic version of tabs, without badges.

Then you can create a new preview, `PreviewBadge.vue`, that demonstrates the `badge` prop as you've done here. To make this work, you'll also need an `ExampleBadge.vue` file. You should be able to copy `ExampleTab.vue` and just modify a few things.

Finally, you'll need to edit `ComponentTabs.vue` to make sure your new example is visible in the list of tab examples on the right.

NateWr:
Oh, and there was one more minor thing. When a `<badge>` is nested inside of a tab, let's contract the right and left padding a bit. This should look alright since we're usually going to be dealing with numbers that aren't too big (`999` or less).

So in `Tabs.vue`, add some CSS like this:

```css
.pkpTabs__button .pkpBadge {
  padding-left: 0.5em;
  padding-right: 0.5em;
}
```

thinkbulecount2:
Hi @NateWr, did the changes you outlined above, let me know if there's anything else missing!!

PR #49, opened 2019-10-31
pkp/pkp-lib#4890 Restore tinymce features to FieldRichTextarea

NateWr:
This commit restores some of the custom tinymce features that were
not ported when FieldRichTextarea was created. These include:

- Image uploads, which are now processed by a new API endpoint
- The pkpTags plugin to insert prepared content
- The handling of visual placeholders for the prepared content
- Support for displaying a word count and word limit

### DIFF
--- a/src/components/Form/fields/FieldRichTextarea.vue
+++ b/src/components/Form/fields/FieldRichTextarea.vue
@@ -45,7 +45,8 @@
 			<div class="pkpFormField--richTextarea__toolbar" :id="toolbarId"></div>
 			<editor
 				class="pkpFormField__input pkpFormField--richTextarea__input"
-				v-model="currentValue"
+				v-model="editorValue"
+				ref="editor"
 				:id="controlId"
 				:toolbar="toolbar"
 				:plugins="plugins"
@@ -53,13 +54,26 @@
 				@onFocus="focus"
 				@onBlur="blur"
 			/>
-			<multilingual-progress
-				v-if="isMultilingual && locales.length > 1"
-				:id="multilingualProgressId"
-				:count="multilingualFieldsCompleted"
-				:total="locales.length"
-				:i18n="i18n"
-			/>
+			<div
+				v-if="(isMultilingual && locales.length > 1) || wordLimit"
+				class="pkpFormField--richTextarea__controlFooter"
+			>
+				<multilingual-progress
+					v-if="isMultilingual && locales.length > 1"
+					:id="multilingualProgressId"
+					:count="multilingualFieldsCompleted"
+					:total="locales.length"
+					:i18n="i18n"
+				/>
+				<div v-if="wordLimit" class="pkpFormField--richTextarea__wordLimit">
+					<icon
+						v-if="wordCount > wordLimit"
+						icon="exclamation-triangle"
+						:inline="true"
+					/>
+					{{ __('wordCount', {count: wordCount, limit: wordLimit}) }}
+				</div>
+			</div>
 		</div>
 		<field-error
 			v-if="errors.length"
@@ -72,32 +86,17 @@
 <script>
 import FieldBase from './FieldBase.vue';
 import Editor from '@tinymce/tinymce-vue';
+import Icon from '@/components/Icon/Icon.vue';
+import debounce from 'debounce';
 
 export default {
 	name: 'FieldRichTextarea',
 	extends: FieldBase,
 	components: {
-		Editor
+		Editor,
+		Icon
 	},
 	props: {
-		size: {
-			type: String,
-			default: 'default'
-		},
-		// @see https://www.tinymce.com/docs/configure/editor-appearance/#toolbar
-		toolbar: {
-			type: String,
-			default() {
-				return 'bold italic superscript subscript | link';
-			}
-		},
-		// @see https://www.tiny.cloud/docs/configure/integration-and-setup/#plugins
-		plugins: {
-			type: String,
-			default() {
-				return 'paste,link';
-			}
-		},
 		// @see https://www.tiny.cloud/docs/configure/integration-and-setup/
 		init: {
 			type: Object,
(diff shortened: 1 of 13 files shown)


src/components/Form/fields/FieldRichTextarea.vue
```diff
@@ -155,11 +281,85 @@ export default {
...
+		 * Get a placeholder for prepared content
+		 *
+		 * @param string key The property name
+		 * @param string value The rendered content value
+		 * @return string
 		 */
-		blur() {
-			this.isFocused = false;
+		getPlaceholder(key, value) {
+			return `<span class="pkpTag mceNonEditable" data-symbolic="${key}">${value}</span>`;
```
asmecher: Missing escaping for `${key}` and especially `{$value}`?

NateWr: That's a good question. Are we talking about a possible vulnerability or just a bug? In terms of a vulnerability, I am struggling to see how the prepared content could be passed into the field control except through, say, a plugin or something. In which case the malicious plugin already has access. Is there a vulnerability here that I'm missing?/

In terms of bugs: yes, it's possible to "break" the implementation by passing a key like `test">something here`. But it seems unlikely, since the key must match the `{$variableName}` syntax that will be used to perform the replacement (both in the browser and on the server side). It strikes me as a coder's mistake rather than something we should protect against. And it doesn't break down everything, just the placeholders.

As it stands, `value` can contain HTML code which I would like to preserve if it is not a security risk. That's because it might be possible to insert a block, like the mailing address, which includes some HTML.

```
123 Anywhere Street<br>
Vancouver, CA
```

Again, as far as I can tell, these placeholders, keys and values, will be coming from plugins or other trusted code. But I could definitely be missing something...

asmecher: Just a matter of consistent hygiene rather than a specific vulnerability in mind. Keeping consistent means we don't have to think through a setting's entire convoluted data path to be sure it's trustworthy.

Values can well be untrusted user data, e.g. given names. Values can be either legitimately HTML (e.g. signature fields) or plaintext (possibly including `<>` characters that are expected to be escaped when presented in an HTML context).

Reviewing how we currently handle this, see https://github.com/pkp/pkp-lib/blob/master/classes/mail/SubmissionMailTemplate.inc.php#L52..L60 for an example. When plaintext content (which may contain `<>` etc.) is passed into an email template, it should be escaped e.g. using `htmlspecialchars` (not sure why we're using `strip_tags` in the above example -- a candidate for fixing, IMO) when it's provided to the email template. Emails are always composed in HTML, so it makes sense to continue to have code providing data to email templates meet that expectation.

Reviewing that, the `${value}` usage above is good, but I'd still prefer (strictly for hygiene purposes) an escape on `${key}`. You're correct that it's a controlled playground and only code sets the keys, so it's academic.

NateWr: Thanks, that's helpful. I can apply an alphanumeric restriction to the key, since I think that would apply to how we format `{$variablesForReplacement}`.

Now that you mention names (and submission titles!), I think we do need to escape the `<>` characters at the very least. That will kind of suck for `<sup>` use cases, but an unintentionally escaped `<sup>` is easier to recover from than a broken `1 < 3` case.

NateWr: https://github.com/pkp/ui-library/commit/e92278b114ab0b0dce847b98d1d9e96457adcd46

asmecher: I don't think we'll be able to reliably pass all content through a single pipe regardless of content type and try to guess the escaping approach from the content. Where data is stored, we *always* know whether it's plain-text or HTML and we'll need to use that.

Consider the "worst case scenario" -- a journal that discusses HTML. Titles are currently expected to be plain-text only, so a submission titled "The use of &lt;blink&gt; in scholarly articles" could never be distinguished from e.g. an abstract (where HTML is allowed) without the pipeline being either 1) informed about what kind of content it's getting, or 2) accepting everything in its input as properly-escaped HTML and treating it accordingly.

The above `SubmissionMailTemplate` example takes option 2. Email templates always expect to work in HTML, so when setting parameters, HTML content is filtered (e.g. `stripUnsafeHtml`) and plain-text is escaped.

NateWr: Ok, so are you suggesting that we shouldn't do any escaping one the client side and rely entirely on escaping on the server side? I can see that working, but one of the goals I had was to support rendering variables in email templates, for example during the review workflow.

So when an email template is selected, the submission title might be merged in from the data that's already loaded in the browser. That data will be equivalent to the data from the REST API about the submission, so won't be pre-escaped.

In such cases, the data isn't specifically delivered to the component from the server. The data is delivered in a general way, through the REST API, so some kind of in-browser escaping would need to happen.

We could do that escaping when the form data is being spun up -- so when the reviewer assignment form loads and the submission details are being passed to the email template, we can escape in-browser then. Does that sound better?

asmecher: Escaping should always be done at the point where the content is being used in another format. As a simple motivating example, it is better to do...

```
$templateMgr->assign('data', array('this' => $needsEscaping));
$templateMgr->display('some.tpl');
```

...and...

```
<script>
var dataFromPhp = {$data|json_encode};
$element.html($('<div/>').text(dataFromPhp.needsEscaping).html());
</script>
```

...rather than...

```
$templateMgr->assign('data', array('this' => htmlspecialchars($needsEscaping)));
```

...and...

```
<script>
var dataFromPhp = {$data|json_encode};
$element.html(dataFromPhp.this);
</script>
```

Maintaining the latter quickly gets into a spaghetti-tracing game and it's really easy for something to slip through. This necessarily means that some escaping gets done client-side and some gets done server-side, depending on the circumstances.

(Forgive the JQueryish code and simple example...)

NateWr: :+1: I agree. If I understand you correctly, this is how it's done now, with the content escaped as its passed to the tinymce control.

asmecher: Excellent, thanks!

PR #77, opened 2020-03-04
Update syntax for draggable options

NateWr:


NateWr:
This collects the following PRs into one so they can be tested together:

https://github.com/pkp/ui-library/pull/74
https://github.com/pkp/ui-library/pull/73
https://github.com/pkp/ui-library/pull/72

PR #84, opened 2020-04-05
Added read only / disabled flag to some components

ViRuSTriNiTy:
see https://github.com/pkp/pkp-lib/issues/5705

  - FieldAutosuggest, FieldText, FieldTextarea, added read only flag
  - FieldSelect, added disabled flag

### DIFF
--- a/src/components/Form/fields/FieldAutosuggest.vue
+++ b/src/components/Form/fields/FieldAutosuggest.vue
@@ -153,7 +153,8 @@ export default {
 			default() {
 				return '';
 			}
-		}
+		},
+		isReadOnly: Boolean
 	},
 	data() {
 		return {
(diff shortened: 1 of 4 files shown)


NateWr:
Also, I'd like to see us standardize on `isDisabled` and `isReadonly`. They can have the same CSS styles applied to them, but apply the different `disabled` and `readonly` HTML attributes. If I recall correctly, the `readonly` attribute allows someone to interact with a field (like select text) but not change it, whereas the `disabled` attribute prevents someone from selecting text at all.

asmecher:
(One important distinction between `disabled` and `readonly` is the user's ability to scroll through long content to read it. That's not possible with a `disabled` field. So in cases like a metadata form that can't currently be edited but is still viewable/readable, `readonly` is a less frustrating user experience than `disabled`.)

PR #101, opened 2020-06-17
i5982 editor filter

thinkbulecount2:
From this issue: https://github.com/pkp/pkp-lib/issues/5982


src/docs/components/Filter/readme.md
```diff
@@ -7,6 +7,22 @@
...
+| `useStars` | Whether or not to display the value as a star rating. Default: `false`. |
+| `valueLabel` | How to display the value. A label of `At least {$value}` will display as `At least 20`. Default: `{$value}`. |
+
+## Props (FilterSliderMultirange)
+
+| Key | Description |
+| --- | --- |
+| `lessThanLabel` | An accessible label for the higher value. Typically "Less than". Required. |
+| `moreThanLabel` | An accessible label for the lower value. Typically "Higher than". Required. |
+
```
NateWr, line 25: Can you also document the props specific to `FilterAutosuggest`? There's only one prop, `autosuggestProps`, and the description for this can just say something like "Props to be passed to the `FieldAutosuggest` component."

thinkbulecount2:
Hi @NateWr updated this PR!

NateWr:
It looks like the `activeFilters` object is not getting loaded correctly in `SubmissionsListPanel.vue`. As a result, the request is not being sent out with the correct form. After adding a couple of section editors with the filter autosuggest, the active filters object looked like this:

active-filters

But it should be a flat array with just the user ids, like this:

active-filters-correct

This is happening because the `addFilter` method in `SubmissionsListPanel` has some logic to handle different filters differently:

https://github.com/pkp/ui-library/blob/master/src/components/ListPanel/submissions/SubmissionsListPanel.vue#L196-L217

In particular, in this section it is handling two different types of filters. 

```js
if (['isOverdue', 'daysInactive'].includes(param)) {
	newFilters[param] = value;
} else {
	if (!newFilters[param]) {
		newFilters[param] = [];
	}
	newFilters[param].push(value);
}
```

This replaces the value with a new value for the `isOverdue` and `daysInactive` filters. But for the other filters it pushes them onto the array.

The `PreviewFilterAutosuggest` shows us how the `add-filter` event should be handled on this line:

https://github.com/pkp/ui-library/pull/101/files#diff-0d6da2ae106ab6ad336dc300e25fa414R41

So to get the functionality we want, we need to make sure that when `addFilter()` is called in `SubmissionsListPanel` with the `assignedTo` parameter, it must be added to the `newFilters` object the same as it is in `PreviewFilterAutosuggest`.

thinkbulecount2:
@NateWr made some changes to `SubmissionsListPanel.vue` and checked that it is showing a flat array:

Screen Shot 2020-06-29 at 09 44 28

src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -206,7 +206,7 @@ export default {
 				if (newFilters.hasOwnProperty('isIncomplete')) {
 					delete newFilters.isIncomplete;
 				}
-				if (['isOverdue', 'daysInactive'].includes(param)) {
+				if (['isOverdue', 'daysInactive', 'assignedTo'].includes(param)) {
```
NateWr, line 209: :boom:

PR #105, opened 2020-07-03
Description for the custom radio input field

Vitaliy-1:


src/docs/components/Form/fields/FieldRadioInput/readme.md
```diff
@@ -4,6 +4,7 @@
 | --- | --- |
 | `...` | Supports all props in [FieldBase](#/component/Form/fields/FieldBase). |
 | `value` | The current value for this field. |
+| `isInput` | Should be set to `true` for a custom field. |
```
NateWr, line 7: This isn't a top-level `prop` so this documentation could be misleading. I think where we need to change the documentation is something like this:

```md
| `...` | Supports all props in [FieldBase](#/component/Form/fields/FieldBase). |
| `...` | Supports all props in [FieldOptions](#/component/Form/fields/FieldOptions). |
| `options` | A list of the options as described in [FieldOptions](#/component/Form/fields/FieldOptions), with additional support for an input field. See [Custom Input](#custom-input).
```

Then in the `##Custom Input` section below, let's provide some sample code:

```md
## Custom input

When creating a field for a custom input, instead of `value` prop, `isInput` should be passed with a boolean value `true`.

```js
[
	{
		value: 1,
		label: "One"
	},
	{
		value: 2,
		label: "Two"
	},
	{
		isInput: true
	}
]
```

NateWr: And since the `value` is a key inside of the `options` prop, let's adjust the terminology slightly:

> When creating an option with a custom input field, pass the `isInput` key instead of the `value` key.

NateWr:
Thanks!

PR #106, opened 2020-07-06
pkp/pkp-lib#5702 Ability to disable submissions in OJS3

salmanm2003:
Hi,
These changes are related to disabling submissions feature.
I have added two changes here:
1- Ability to show/hide buttons.
2- Fixed the issue of having space underneath the notifications on submissions list page.

Thanks 
Salman


src/components/Container/SettingsPage.vue
```diff
@@ -74,7 +74,7 @@ export default {
 <style lang="less">
 @import '../../styles/_import';
 
-.pkp_page_management .app__page > .pkpNotification {
+.pkp_page_management .app__page > .pkpNotification, .pkp_page_submissions .app__page > .pkpNotification  {
```
NateWr, line 77: This looks good, however, it mixes responsibilities. The `SettingsPage` should only have styles related to the settings page. Since there isn't a page component specific to submissions, you can add this style to `Page.vue`.

Or, I'd suggest getting rid of the notification altogether, as I mentioned in a comment on the pkp-lib PR. If you do that, that would resolve this issue anyway.

src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -22,6 +22,7 @@
 					</pkp-button>
 					<pkp-button
 						v-if="addUrl && currentUserCanAddSubmission"
+						v-show="showButton"
```
NateWr, line 25: Let's use the existing `v-if` and `currentUserCanAddSubmission` computed property. In `SubmissionsListPanel.vue` you'll find:

```
	currentUserCanAddSubmission() {
		return this.userHasRole([
			...
		]);
	}
```

Let's add the global disable flag there:

```
	currentUserCanAddSubmission() {
		return this.showButton && this.userHasRole([
			...
		]);
	}
```

src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -154,6 +155,12 @@ export default {
 		title: {
 			type: String,
 			required: true
+		},
+		showButton: {
```
NateWr, line 159: We may have other buttons in this panel in the future, so let's call this something more descriptive like `allowSubmissions`.

src/components/ListPanel/submissions/SubmissionsListPanel.vue
```diff
@@ -22,6 +22,7 @@
 					</pkp-button>
 					<pkp-button
 						v-if="addUrl && currentUserCanAddSubmission"
+						v-show="allowSubmissions"
```
NateWr, line 25: This line can be removed now because the display condition should be handled by `currentUserCanAddSubmission` in the line above.

PR #120, opened 2020-11-04
pkp/pkp-lib#2615 Only show delete button for declined submissions

NateWr:


luizbgomide:
Thanks a lot, this will be very helpful and will avoid accidental deletions.

luizbgomide:
@NateWr, how do I apply this patch?

NateWr:
@luizbgomide you'll need to rebuild the JavaScript package in order to apply this change right away. You can also wait until 3.3 is released and the fix will be applied when you upgrade to 3.3.

luizbgomide:
Thanks, I will try to rebuild the package. I've never worked with Javascript before but I suppose that I just need to run the following on my git source, right?

```
npm install
npm run build
```

NateWr:
Yes, after applying the change that should do it.

NateWr:
The updated package is built and placed in `js/build.js`.

PR #123, opened 2020-11-15
Handle smarty variables in Tinymce

withanage:
@NateWr 

I have added the  logic to handle smarty variables into tinymce using a callback function.

This jsfiddle shows the functionality for it. 
https://fiddle.tiny.cloud/xyhaab

Issue [#122](https://github.com/pkp/orcidProfile/issues/122)

In SiteHandler.js there is also a [place](https://github.com/pkp/pkp-lib/blob/master/js/controllers/SiteHandler.js#L167), where tinymce is used , let me know, if that is still used  and the logic has to be added there too.

A small demo of the added functionality.
tinymce

### DIFF
--- a/src/components/Form/fields/FieldRichTextarea.vue
+++ b/src/components/Form/fields/FieldRichTextarea.vue
@@ -178,6 +178,11 @@ export default {
 				relative_urls: false,
 				remove_script_host: false,
 				convert_urls: true,
+				urlconverter_callback: function (url) {
+					// removes script_host from smarty variables
+					var smartyVariable = /\{\$(\w*)\}/.exec(url);
+					return (smartyVariable) ? smartyVariable[0] : window.location.href + '/' + url;
+				},
 				// See: https://www.tiny.cloud/docs/general-configuration-guide/upload-images/#rollingyourimagehandler
 				images_upload_handler(blobInfo, success, failure) {
 					const data = new FormData();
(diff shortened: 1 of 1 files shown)


src/components/Form/fields/FieldRichTextarea.vue
```diff
@@ -178,6 +178,14 @@ export default {
...
 				remove_script_host: false,
 				convert_urls: true,
+				urlconverter_callback: function (url) {
+					// removes script_host from smarty variables
+					var absoluteUrl = /^http[s]*:\/\//i.test(url);
+					var smartyVariable = /\{\$(\w*)\}/.exec(url);
+					if (!absoluteUrl)  url = window.location.href + '/' + url;
+					if (smartyVariable) url = smartyVariable[0] ;
+					return  url;
+				},
```
NateWr, line 188: Use the new ES6 syntax for consistency:

```
urlconverter_callback(url) {
```

And the use `const` instead of `var` for these.

Also, there's an extra space between `return` and `url` on line 187.

NateWr (changes requested):
Looks good! Go ahead and port it to `master` as well.

src/components/Form/fields/FieldRichTextarea.vue
xmlFlow: @NateWr 
Thanks for the Tipp with the ui-library linter. 
I have added the changes now.

withanage: @NateWr
Thanks for the Tipp with the ui-library linter.
I have added the changes now.

PR #165, opened 2021-09-29
pkp/pkp-lib#7014 Bring DOIs into the core application and refactor to support deposit status

ewhanson:


src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+			if (['status', 'isPublished', 'issueIds'].includes(param)) {
+				// Handle "toggleable" filters
+				delete newFilters[param];
+			} else {
+				// Handle multi-select filters
+				newFilters[param] = newFilters[param].filter(v => v !== value);
+			}
+			this.activeFilters = newFilters;
+		},
+		openDepositDialog(itemIds = [], action = 'deposit') {
```
NateWr, line 370: Although it is right from a DRY perspective to try to reuse code in this way, I think in time you'll find that the abstraction of passing an action to the `openDepositDialog` and `executeExportAction` and `onExport...` methods will be more work than its worth.

I'd suggest using distinct methods to make the code more readable and extensible over time, and using `this.selected` instead of passing in the items:

```
openBulkDeposit() {
   const count = this.selected.length;
   this.openDialog({
     ....
   });
},
openBulkExport() {
   this.openDialog({
     ....
   });
},
openBulkRegister() {
   this.openDialog({
     ....
   });
},
```

When you want to reuse code that is just configuration, I recommend composing the configuration parameters like this:

```
getBulkActionAjaxProps(action) {
  return {
    url: `${this.executeActionApiUrl}/${action}`,
    // ... everything else, but not the success/complete methods
  };
},
openBulkDeposit() {
  this.openDialog({
    ...
    callback: () => {
      $ajax({
        ...this.getBulkActionAjaxProps('dialog'),
        data: {ids: this.selected},
        success(r) {
          // do stuff
        },
    });
  });
}
```

With this approach (sometimes called "composition"), you abstract some of the reusable configuration, but in a way that allows you to recompose your own configuration on top of it. That prevents too much code from being coupled tightly together. If one action needs special handling of a success/error condition, you can write that in without disrupting the others.

And on the plus side, you end up with documented methods for each action, which makes the templates easier to read too.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+		 * Callback to fire when the form submission's ajax request has been
+		 * returned successfully
+		 *
+		 * @param {Object} response The response to the AJAX request
+		 * @param {String} textStatus Response status as text
+		 * @param {Object} jqXHR jQuery superset of XMLHttpRequest
+		 * @param {String} action The export action executed (deposit, export, markRegistered)
+		 */
+		onExportSuccess(response, textStatus, jqXHR, action) {
+			window.console.log('[Success]:', response);
```
NateWr, line 439: When you're ready to deal with success and error messages, have a look at [Notify](https://docs.pkp.sfu.ca/dev/ui-library/dev/#/utilities/Notify) for success messages and `ajaxErrorCallback` for error messages.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+				newItem = this.addDoiObjects(newItem);
+
+				return newItem;
+			});
+		}
+	},
+	mounted() {
+		this.$on('deposit-triggered', (id, action) => {
+			this.openDepositDialog([id], action);
+		});
```
NateWr, line 621: This shouldn't be needed anymore, because you're listening to the event from `DoiListItem` with the `@deposit-triggered` line.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+		line-height: 1.5rem;
+
+		> input {
+			margin-left: 0.5rem;
+		}
+	}
+
+	.listPanel__selectAllLabel {
+		margin-left: 0.5rem;
+	}
```
NateWr, line 646: These first two blocks related to select all are no longer used I think.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+		display: flex;
+		align-items: center;
+		margin-left: -1rem;
+	}
+
+	.listPanel__selector {
+		line-height: 100%;
+		width: 3rem;
+		padding-left: 1rem;
+	}
```
NateWr, line 658: I'd rename these classes to be specific to the DOI list panel, since they're not part of the `listPanel` code itself.

NateWr (commented):
Looks good. I'm going on to the pkp-lib PR now.

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -0,0 +1,754 @@
...
+			});
+
+			this.mutableDois = dois;
+		},
+		editOrSaveDois(shouldSave) {
+			shouldSave ? this.saveDois() : this.editDois();
+		},
+		editDois() {
+			this.isEditingDois = true;
+			this.isEditingDoisEnabled = true;
```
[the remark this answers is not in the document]
ewhanson: `isEditingDois` was tracking when the entire editing processing including saving all individual DOIs was completed, and `isEditingDoisEnabled` was for when the `FieldDoiText`'s text box had editing enabled. I think the idea here was that the ability to change what was in the text box was disabled while all the DOIs were saving but we still need to know when the process as a whole was complete. Would something like `isEditingTextboxEnabled` work better?

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -0,0 +1,669 @@
...
+		 * @param {Boolean} expand Whether it should be expanded or closed
+		 */
+		expandItem(itemId, expand) {
+			const expanded = this.expanded.includes(itemId);
+			if (expand && !expanded) {
+				this.expanded.push(itemId);
+			} else if (!expand && expanded) {
+				this.expanded = this.expanded.filter(item => item !== itemId);
+			}
+		},
```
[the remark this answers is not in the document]
ewhanson: Did this for toggling selected as well.

src/components/ListPanel/doi/DoiListItem.vue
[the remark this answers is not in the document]
NateWr: Sorry I'm still not getting it. As far as I can tell these are the points at which state changes:

- Click edit button = () => this.isEditingDois = true
- Click save button = () => this.isEditingDois = false; this.isSaving = true;
- Save complete = () => this.isSaving = false;

Does that cover it? Is `this.isEditingDoisEnabled` the same as `this.isSaving`?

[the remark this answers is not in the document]
ewhanson: Thanks, that makes sense. It was the equivalent of `isEditingDois && !isSaving` so I'm using that instead. The way it currently is, saving happens within the editing process so`isEditingDois` will still be true until all saving actions have been completed.

NateWr:
Merged manually.

PR #171, opened 2021-12-06
pkp/pkp-lib#6675 Added extra features to the Autosuggest field (showP…

jonasraoni:
…agination, minInputLength, maxSelectedItems, replaceWhenFull) and a new component (FieldPagedAutosuggest) to enter a custom data mapper to feed the suggestions model.

### DIFF
--- a/src/components/Filter/FilterAutosuggest.vue
+++ b/src/components/Filter/FilterAutosuggest.vue
@@ -7,12 +7,14 @@
 <script>
 import Filter from './Filter.vue';
 import FieldAutosuggestPreset from '@/components/Form/fields/FieldAutosuggestPreset.vue';
+import FieldPagedAutosuggest from '@/components/Form/fields/FieldPagedAutosuggest.vue';
 import FieldSelectUsers from '@/components/Form/fields/FieldSelectUsers.vue';
 import FieldSelectIssues from '@/components/Form/fields/FieldSelectIssues.vue';
 export default {
 	extends: Filter,
 	components: {
 		FieldAutosuggestPreset,
+		FieldPagedAutosuggest,
 		FieldSelectUsers,
 		FieldSelectIssues
 	},
--- a/src/components/Form/fields/FieldBaseAutosuggest.vue
+++ b/src/components/Form/fields/FieldBaseAutosuggest.vue
@@ -66,7 +66,26 @@
 				class="pkpFormField--autosuggest__autosuggest"
 				v-bind="autosuggestOptions"
 				@selected="selectSuggestion"
-			/>
+			>
+				<template slot="after-suggestions">
+					<div
+						class="pkpFormField--autosuggest__pagination"
+						v-if="showPagination && lastPage > 1"
+						@click.prevent.stop=""
+						@mouseup.prevent.stop=""
+						@mousedown.prevent.stop=""
+					>
+						<pagination
+							@set-page="setPage"
+							v-bind="paginationOptions"
+							:currentPage="currentPage"
+							:isLoading="isLoading"
+							:lastPage="lastPage"
+						/>
+					</div>
+				</template>
+			</vue-autosuggest>
+			<spinner class="pkpFormField--autosuggest__spinner" v-if="isLoading" />
 			<div
 				v-if="currentPosition === 'below'"
 				class="pkpFormField--autosuggest__values pkpFormField--autosuggest__values--below"
@@ -108,35 +127,25 @@
 <script>
 import FieldBase from './FieldBase.vue';
 import PkpBadge from '@/components/Badge/Badge.vue';
+import Pagination from '@/components/Pagination/Pagination.vue';
 import {VueAutosuggest} from 'vue-autosuggest';
-import ajaxError from '@/mixins/ajaxError';
 import debounce from 'debounce';
+import fetch from '@/mixins/fetch';
 import elementResizeEvent from 'element-resize-event';
 
 export default {
 	name: 'FieldBaseAutosuggest',
 	extends: FieldBase,
-	mixins: [ajaxError],
+	mixins: [fetch],
 	components: {
 		PkpBadge,
-		VueAutosuggest
+		VueAutosuggest,
+		Pagination
 	},
 	props: {
-		apiUrl: {
-			type: String,
-			default() {
-				return '';
-			}
-		},
 		deselectLabel: {
 			type: String,
 			required: true
-		},
-		getParams: {
-			type: Object,
-			default() {
-				return {};
-			}
 		},
 		initialPosition: {
 			type: String,
@@ -156,6 +165,22 @@ export default {
 		selectedLabel: {
 			type: String,
 			required: true
+		},
+		showPagination: {
+			type: Boolean,
+			default: true
+		},
+		minInputLength: {
+			type: Number,
+			default: 0
+		},
+		maxSelectedItems: {
+			type: Number,
+			default: null
+		},
+		replaceWhenFull: {
+			type: Boolean,
+			default: true
 		}
 	},
 	data() {
@@ -257,6 +282,17 @@ export default {
 				id: this.controlId,
 				name: this.name
 			};
+		},
+
+		/**
+		 * Options to pass to the pagination component
+		 *
+		 * @return {Object}
+		 */
+		paginationOptions() {
+			return {
+				showAdjacentPages: 2
+			};
 		}
 	},
 	methods: {
@@ -275,29 +311,30 @@ export default {
 		},
 
 		/**
-		 * Get suggestions from the API url
+		 * Update the list of items
+		 *
+		 * @param {Array} items
+		 * @param {Number} itemsMax
 		 */
-		getSuggestions: debounce(function() {
+		setItems(items, itemsMax) {
+			this.setSuggestions(items, itemsMax);
+		},
+
+		/**
+		 * Overwrite the get method from the fetch mixin to avoid requesting a resource if there's no user input
+		 */
+		get() {
 			if (!this.inputValue) {
 				this.suggestions = [];
 				return;
 			}
-			var self = this;
-			$.ajax({
-				url: this.apiUrl,
-				type: 'GET',
-				data: {
-					...this.getParams,
-					searchPhrase: this.inputValue
-				},
-				error(r) {
-					self.ajaxErrorCallback(r);
-				},
-				success(r) {
-					self.setSuggestions(r.items);
-				}
-			});
-		}, 250),
+			this.getSuggestions();
+		},
+
+		/**
+		 * Get suggestions from the API url by leveraging the get method of the fetch mixin
+		 */
+		getSuggestions: debounce(fetch.methods.get, 250),
 
 		/**
 		 * Add a suggested item to the list of selected items
@@ -316,7 +353,21 @@ export default {
 				}
 				item = this.suggestions[0];
 			}
-			this.currentSelected.push(item);
+			if (!this.currentSelected.find(({value}) => value === item.value)) {
+				const hasReachedLimit =
+					this.maxSelectedItems !== null &&
+					this.currentSelected.length >= this.maxSelectedItems;
+				const canReplace =
+					hasReachedLimit &&
+					this.currentSelected.length &&
+					this.replaceWhenFull;
+				if (!hasReachedLimit || canReplace) {
+					if (canReplace) {
+						this.currentSelected.pop();
+					}
+					this.currentSelected.push(item);
+				}
+			}
 			this.inputValue = '';
 			this.$nextTick(() => {
 				this.$nextTick(() => {
@@ -342,9 +393,10 @@ export default {
 		 * This must be implemented in a component that extends
 		 * this component
 		 *
-		 * @param {Array} newItems
+		 * @param {Array} items
+		 * @param {Number} itemsMax
 		 */
-		setSuggestions(newItems) {
+		setSuggestions(items, itemsMax) {
 			throw new Error(
 				'The setSuggestions method must be implemented in any component that extends FieldBaseAutosuggest.'
 			);
@@ -404,7 +456,9 @@ export default {
 			if (newVal === oldVal) {
 				return;
 			}
-			this.getSuggestions();
+			if (newVal.length >= this.minInputLength) {
+				this.setSearchPhrase(newVal);
+			}
 		}
 	},
 	mounted() {
@@ -483,6 +537,17 @@ export default {
 
 .pkpFormField--autosuggest__autosuggest {
 	position: relative;
+}
+
+.pkpFormField--autosuggest__pagination {
+	border-top: @bg-border;
+	padding: 0.5rem 1rem;
+}
+
+.pkpFormField--autosuggest__spinner {
+	position: absolute;
+	right: 1rem;
+	top: 0.2rem;
 }
 
 .pkpFormField--autosuggest__input {
--- a/src/components/Form/fields/FieldControlledVocab.vue
+++ b/src/components/Form/fields/FieldControlledVocab.vue
@@ -30,12 +30,14 @@ export default {
 						self.ajaxErrorCallback(r);
 					},
 					success(r) {
-						self.allSuggestions = r.map(v => {
-							return {
-								value: v,
-								label: v
-							};
-						});
+						self.allSuggestions = r
+							.map(v => {
+								return {
+									value: v,
+									label: v
+								};
+							})
+							.sort();
 						self.setSuggestions.apply(self);
 						self.suggestionsLoaded = true;
 					}
@@ -80,6 +82,9 @@ export default {
 					(this.inputValue !== suggestion.value &&
 						suggestion.value.match(regex))
 			);
+			this.itemsMax = this.suggestions.length;
+			const start = (this.currentPage - 1) * this.count;
+			this.suggestions = this.suggestions.slice(start, start + this.count);
 			if (this.inputValue && !this.suggestions.includes(this.inputValue)) {
 				this.suggestions.unshift({
 					value: this.inputValue,
--- a/src/components/Form/fields/FieldPagedAutosuggest.vue
+++ b/src/components/Form/fields/FieldPagedAutosuggest.vue
@@ -1,0 +1,25 @@
+<script>
+import FieldBaseAutosuggest from './FieldBaseAutosuggest.vue';
+
+export default {
+	name: 'FieldPagedAutosuggest',
+	extends: FieldBaseAutosuggest,
+	props: {
+		dataMapper: {
+			type: Function
+		}
+	},
+	methods: {
+		/**
+		 * Set the suggestions
+		 *
+		 * @param {Array} items List of items
+		 * @param {Number} itemsMax Total amount of items from the API
+		 */
+		setSuggestions(items, itemsMax) {
+			this.suggestions = this.dataMapper ? items.map(this.dataMapper) : items;
+			this.itemsMax = itemsMax;
+		}
+	}
+};
+</script>
--- a/src/components/Form/fields/FieldSelectIssues.vue
+++ b/src/components/Form/fields/FieldSelectIssues.vue
@@ -8,23 +8,21 @@ export default {
 		/**
 		 * Set the suggestions from an API response
 		 *
-		 * Maps the API response when searching for submissions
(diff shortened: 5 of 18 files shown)


src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -66,7 +66,26 @@
 				class="pkpFormField--autosuggest__autosuggest"
 				v-bind="autosuggestOptions"
 				@selected="selectSuggestion"
-			/>
+			>
+				<template slot="after-suggestions">
+					<div
+						class="pkpFormField--autosuggest__pagination"
+						v-if="showPagination && lastPage > 1"
```
jonasraoni, line 73: If there's no pages, the pagination will be hidden, I'm not sure that addresses the WCAG issues...

As an user I had bad experiences with limited lists, so I'd like to see everything, sometimes I don't know what to search very well.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -66,7 +66,26 @@
...
+						@click.prevent.stop=""
+						@mouseup.prevent.stop=""
+						@mousedown.prevent.stop=""
```
jonasraoni, line 76: Just to prevent the popup from closing before the pagination is triggered.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -66,7 +66,26 @@
...
+							@set-page="setPage"
+							v-bind="paginationOptions"
+							:currentPage="currentPage"
+							:isLoading="isLoading"
+							:lastPage="lastPage"
+						/>
+					</div>
+				</template>
+			</vue-autosuggest>
+			<spinner class="pkpFormField--autosuggest__spinner" v-if="isLoading" />
```
jonasraoni, line 88: I thought the fields were simple text fields due to the slow response, so I added a loading here.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -275,29 +311,30 @@ export default {
...
 
 		/**
-		 * Get suggestions from the API url
+		 * Update the list of items
+		 *
+		 * @param {Array} items
+		 * @param {Number} itemsMax
+		 */
+		setItems(items, itemsMax) {
+			this.setSuggestions(items, itemsMax);
```
jonasraoni, line 320: I thought about getting rid of the `getSuggestions`, `setSuggestions`, etc., given that the fetch already have useful methods, but I wanted to confirm whether you're ok with it.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -316,7 +353,21 @@ export default {
 				}
 				item = this.suggestions[0];
 			}
-			this.currentSelected.push(item);
+			if (!this.currentSelected.find(({value}) => value === item.value)) {
```
jonasraoni, line 356: I saw the sub-components had a filter to avoid listing an already selected item, so as a defensive approach I've moved it here. Given that the `item.value` is used as a `:key`, we can't have duplicates here anyway.

But better to keep both =]

src/components/Form/fields/FieldControlledVocab.vue
```diff
@@ -80,6 +82,9 @@ export default {
 					(this.inputValue !== suggestion.value &&
 						suggestion.value.match(regex))
 			);
+			this.itemsMax = this.suggestions.length;
+			const start = (this.currentPage - 1) * this.count;
+			this.suggestions = this.suggestions.slice(start, start + this.count);
```
jonasraoni, line 87: I've added the sort and also this piece to decrease the long list issue =]

src/components/Form/fields/FieldPagedAutosuggest.vue
```diff
@@ -0,0 +1,25 @@
...
+		/**
+		 * Set the suggestions
+		 *
+		 * @param {Array} items List of items
+		 * @param {Number} itemsMax Total amount of items from the API
+		 */
+		setSuggestions(items, itemsMax) {
+			this.suggestions = this.dataMapper ? items.map(this.dataMapper) : items;
+			this.itemsMax = itemsMax;
+		}
```
jonasraoni, line 22: Just created this component in order to be able to have custom mappers, in my case I wanted to have a "Full name (email)", another option would be to customize the `FieldSelectUsers` 🤔

NateWr:
I'm really sorry, but I think I'm going to say no to pagination in the autosuggest field. I've had a look at widely used autosuggest patterns and I haven't seen pagination implemented much. I did find a couple of libraries that supported pagination through an infinite scroll technique (ie - load more as they scroll), but this has its own usability problems. Because autosuggest is already a tricky component for accessibility, I'm not willing to commit to more UI complexity in it. And I especially don't want to adopt an unusual pattern that hasn't really been proven in the wild.

The autosuggest patten has a few common conventions and one of them is that you continue to type to refine your search. If the list is so large that this doesn't provide a good UX, then we need to redesign the UI to not use an autosuggest. But I would only want to do that in response to UX research that showed this was a problem.

I can suggest an alternate way forward here. First, we should cap the height of the suggestions container and let it scroll, so that a large number of items in the list doesn't require scrolling away from the input field. See this [example](https://darrenjennings.github.io/vue-autosuggest/storybook/?selectedKind=Vue-Autosuggest&selectedStory=many%20many%20options&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) (search for numbers). Then, when more than one page of suggestions exists, the component can continue to make requests for other pages in the background and add them to the list as they come in. So the list of suggestions would grow as subsequent pages were loaded in.

Your addition of the loading spinner is great. And that will help here too. The spinner should remain visible as subsequent pages are loaded in. I suspect we'll have some accessibility issues around this, but it will likely be the same with or without the spinner.

The one trick here is that you'll want to make sure that you bail when the user's search changes. You don't want to be gathering pages and pages of results for an obsolete search phrase.

NateWr:
Another approach which I've considered, but haven't yet researched, is to provide a "Load More" button at the end of the results. The autosuggest component has a [footer](https://github.com/darrenjennings/vue-autosuggest#headerfooter) slot that might be used for this.

autosuggest-load-more

The trick is that we'll need to be careful about how we use the aria attributes to link the button to the component. My first thought is that we'll need an `aria-controls` attribute, but I'm not 100% sure.

jonasraoni:
No worries, I just wanted to close the issue fast with a generic solution 😁

I'm not able to tell which option is more accessible, I'd personally opt-in for the infinite scroll (the less I click, the better), as I'll be able to keep typing and scroll at the same time.

Then about loading, I'd load 2-3 pages in advance and wait the user to reach the middle/end of the list before loading more.

jonasraoni:
This site is an old pearl! About the frustration I think people must be already used to it, given the amount of doom scrollers out there 😂

Well, as long as we're not loading the whole result set, I'm ok with any UI... So let's go with the "Load more".

jonasraoni:
@NateWr I replaced the pagination by the load more and added a scrollbar.

Note: The `vue-autosuggest` with a scrollbar depends on a [`mouseup`](https://github.com/darrenjennings/vue-autosuggest/blob/master/src/Autosuggest.vue#L379) attached to the `document`, but our `SiteHandler.js` is consuming all `mouseup` events and breaking the bubbling, so I had it updated as well.

public/globals.js
```diff
@@ -89,6 +89,8 @@ window.pkp = {
 		'common.pagination.label': 'View additional pages',
 		'common.pagination.next': 'Next page',
 		'common.pagination.previous': 'Previous page',
+		'common.pagination.loadMore': 'Load more',
+		'common.pagination.loadMore.description': '{$quantity} results not shown',
```
jonasraoni, line 93: Given that our pluralization handling will be incorrect even if we add a `quantity > 1 ? plural : singular`, I've decided to leave this way until the localization framework gets updated.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -482,9 +537,21 @@ export default {
 }
 
 .pkpFormField--autosuggest__autosuggest {
+	--maxAutosuggestHeight: 100%;
```
jonasraoni, line 540: Given that I don't have access to the inner elements to insert a "style: xxx", I decided to use a CSS variable...
But I'm not sure which [browsers we're supporting](https://caniuse.com/css-variables), if that's not welcome, I can leave the value fixed or find another solution (I still have other ideas).

src/components/Form/fields/FieldMappedAutosuggest.vue
```diff
@@ -0,0 +1,30 @@
...
+		 * @param {Array} items List of items
+		 * @param {Number} itemsMax Total amount of items from the API
+		 */
+		setSuggestions(items, itemsMax) {
+			const suggestions = this.dataMapper ? items.map(this.dataMapper) : items;
+			if (this.offset) {
+				this.suggestions.push(...suggestions);
+			} else {
+				this.suggestions = suggestions;
+			}
```
jonasraoni, line 25: I didn't like much this snippet 🤔
I could replace it by a one-liner `this.suggestions.splice(this.offset, this.suggestions.length, ...items)` or implement this in the `FieldBaseAutosuggest` and update the function to work as a mapper/callback (`return { items, itemsMax }`).

src/components/Form/fields/FieldBaseAutosuggest.vue
NateWr: I'll be honest this makes me nervous. Interfering with interaction events like this can often lead to accessibility problems. Is the suggestions box closing because focus is being dropped? In such cases, you may want to set focus to another element, preferably one that indicates what is happening (like a spinner with help text that says "loading more").

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -66,7 +66,31 @@
...
+							@click="setPage(currentPage + 1)"
+							:aria-controls="controlId"
+						>
+							{{ __('common.pagination.loadMore') }}
+						</pkp-button>
+						{{
+							__('common.pagination.loadMore.description', {
+								quantity: this.unloadedItems
+							})
+						}}
```
NateWr, line 89: Add a spinner when loading more.

src/components/Form/fields/FieldBaseAutosuggest.vue
NateWr: Yes, I like how you've used the `fetch` mixin here. You can migrate the methods over.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -316,7 +354,21 @@ export default {
...
+				const canReplace =
+					hasReachedLimit &&
+					this.currentSelected.length &&
+					this.replaceWhenFull;
+				if (!hasReachedLimit || canReplace) {
+					if (canReplace) {
+						this.currentSelected.pop();
+					}
+					this.currentSelected.push(item);
+				}
```
NateWr, line 370: I'd like to get rid of the `canReplace` functionality. I think it will be confusing for users, who may not notice that an existing selection has been replaced.

Instead, when the max count has been reached, set `inputProps.readonly` to true and make sure the suggestions are empty. Then the user will have to delete their existing selection to search for a new one. I know it's a little bit more clicking, but less likely to cause user error this way.

src/components/Form/fields/FieldBaseAutosuggest.vue
NateWr: I'd like to set this in the CSS code using an `em` height. If a parent component wants to override this, it can write a CSS rule to override it in a specific case.

Here's a PR that moves the overflow to the list and touches up a couple small things with edges in the component: https://github.com/jonasraoni/ui-library/pull/2. I moved the overflow to the list so when overflowing, the load more button is always available. Otherwise, when using the keyboard to move down the list, the button never shows up:


https://user-images.githubusercontent.com/2306629/146036785-69f01f86-26e8-40f0-8072-1a4886d67f7d.mp4

By moving the overflow to the list, the Load More button is visible.


https://user-images.githubusercontent.com/2306629/146037688-b02db2f9-cf8f-4ad7-b60c-4647d9b6d131.mp4

However, probably still some work would need to be done so that the list scrolls as the user goes through it with the keyboard (down key). Otherwise, the user's current focus goes out of view inside the scrollable area.

src/components/Form/fields/FieldMappedAutosuggest.vue
```diff
@@ -0,0 +1,30 @@
+<script>
+import FieldBaseAutosuggest from './FieldBaseAutosuggest.vue';
+
+export default {
+	name: 'FieldMappedAutosuggest',
```
NateWr, line 5: Instead of creating a generic component that takes a mapping function as a prop, I'd prefer to use a specific component built for this use case. All of the other autosuggest fields are set up to use `setSuggestions` as the mapping function, and by the time it is compiled it won't add much to the build size to have a separate component for each.

So let's call this `EnrollUserAsReviewerAutosuggest` and move the mapping function into the `setSuggestions` method.

src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -404,7 +457,9 @@ export default {
 			if (newVal === oldVal) {
 				return;
 			}
-			this.getSuggestions();
+			if (newVal.length >= this.minInputLength) {
```
NateWr, line 460: This prevents suggestions from updating when I delete characters. When the input value drops below `minInputLength`, go ahead and empty the suggestions array.

src/components/Form/fields/FieldSelectIssues.vue
```diff
@@ -8,23 +8,25 @@ export default {
...
+					value: item.id,
+					label: item.identification
+				};
+			});
+			if (this.offset) {
+				this.suggestions.push(...suggestions);
+			} else {
+				this.suggestions = suggestions;
+			}
+			this.itemsMax = itemsMax;
```
NateWr, line 29: Rather than extend every autosuggest field with pagination, let's only support it where we need it. We shouldn't need to manage offset or pagination in any other autosuggest field.

src/docs/components/Filter/previews/PreviewFilterAutosuggest.vue
```diff
@@ -36,9 +36,10 @@ export default {
 				...fieldBase,
 				...fieldBaseAutosuggest,
 				name: 'assignedTo',
-				label: 'Assigned to Editors',
+				label: 'Assigned to Editors (limited to 3 items)',
 				selectedLabel: 'Assigned',
-				apiUrl: '/usernames.json'
+				apiUrl: '/usernames.json',
+				maxSelectedItems: 3
```
NateWr, line 42: Set up a new preview component if you want to demonstrate the `maxSelectedItems` prop. You don't need to for merge, though.

src/mixins/fetch.js
```diff
@@ -44,7 +44,9 @@ export default {
 			isLoading: false,
 			latestGetRequest: '',
 			offset: 0,
-			searchPhrase: ''
+			itemsMax: 0,
+			searchPhrase: '',
+			lastRequest: null
```
NateWr, line 49: Good idea moving this to `fetch.js`. We may be able to get rid of some code that checks this in components that use the mixin. :+1:

NateWr (changes requested):
This looks really good, @jonasraoni. I've left some comments throughout, and I'm really happy with how this has come along. However, I've found a real blocker in terms of a11y and I'm not sure how to work around it. Maybe you'll have a better idea.

A key part of the accessibility handling is the markup that relates the `<input>` field to the `role="listbox"`, and implements required keyboard controls (up/down/enter) to keep the interaction accessible. However, the Load More button can't be accessed through these controls. The user must use tab to go to the next thing to reach the button (or use the mouse).

https://user-images.githubusercontent.com/2306629/146040665-2387f101-b87a-448c-b8ea-9f7b5804aca3.mp4

Tab seems easy enough. But as far as I can tell, there's no common, accepted way to indicate to a screen reader that more controls for this list can be accessed with a tab. I've looked at some examples ([gov.uk](https://alphagov.github.io/accessible-autocomplete/examples/), [WAI combobox](https://www.w3.org/TR/wai-aria-practices-1.1/#examples-1)), but I don't see any examples of an accessible approach here.

I know you've put a lot of work into this. But I don't want to add a feature to the autosuggest that I know will not be accessible.

This may be a case where it's better to step back and think about an alternate approach to the UI problem. We may need something more like a ListPanel for enrolling a user as a reviewer. That said, I'm open to exploring this further. I just did some quick research. But if you can find a trustworthy example of implementing a load more button like this in an accessible way, we can pursue it further. I think, aside from this, the implementation was looking good.

src/docs/components/Filter/previews/PreviewFilterAutosuggest.vue
jonasraoni: Ok!

jonasraoni:
Hi @NateWr, I worked a bit on this yesterday and I had success to make the load more accessible through the keyboard.
If you think it's still not acceptable, then I'll go with the other approach ("Showing 20 of 122 suggestions. Keep typing to refine your search").

src/components/Form/fields/FieldControlledVocab.vue
```diff
@@ -30,14 +31,23 @@ export default {
...
-								value: v,
-								label: v
-							};
-						});
-						self.setSuggestions.apply(self);
+						self.allSuggestions = r
+							.sort(
+								new Intl.Collator(
+									(this.localeKey || $.pkp.app.currentLocale).split('_')[0]
+								).compare
```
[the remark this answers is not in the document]
jonasraoni: I agree! I just saw the list wasn't sorted and wanted to address it temporarily, until the https://github.com/pkp/pkp-lib/issues/7529 is addressed :)

src/docs/components/Form/fields/FieldBaseAutosuggest/readme.md
```diff
@@ -26,13 +31,14 @@ import FieldBaseAutosuggest from './FieldBaseAutosuggest.vue';
 export default {
 	extends: FieldBaseAutosuggest,
 	methods: {
-		setSuggestions(newItems) {
-			this.suggestions = newItems.map(item => {
+		setSuggestions(items, itemsMax) {
```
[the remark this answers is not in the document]
jonasraoni: No problem =]

src/components/Filter/FilterAutosuggest.vue
```diff
@@ -7,12 +7,14 @@
...
 import Filter from './Filter.vue';
 import FieldAutosuggestPreset from '@/components/Form/fields/FieldAutosuggestPreset.vue';
+import FieldMappedAutosuggest from '@/components/Form/fields/FieldMappedAutosuggest.vue';
 import FieldSelectUsers from '@/components/Form/fields/FieldSelectUsers.vue';
 import FieldSelectIssues from '@/components/Form/fields/FieldSelectIssues.vue';
 export default {
 	extends: Filter,
 	components: {
 		FieldAutosuggestPreset,
+		FieldMappedAutosuggest,
```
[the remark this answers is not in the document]
jonasraoni: I used in the enroll reviewer interface, but as you asked to create a custom component, this will be removed.

src/components/Form/fields/FieldBaseAutosuggest.vue
jonasraoni: I agree, I thought about renaming it to `singleSelection: Boolean`, then it would address my use case.
Another option would be to show a warning message "Only N items can be selected", but I saw the errors are injected from outside and didn't want to mess with adding extra messages.

src/components/Form/fields/FieldBaseAutosuggest.vue
jonasraoni: I saw this problem, but I forgot to fix 😁

src/components/Form/fields/FieldSelectIssues.vue
jonasraoni: Agreed! I had to cancel the pagination in the controlled vocabs component, but the other way would feel more natural.

jonasraoni:
Note: the "load more" will be at the end of the list, but it's doable to implement your design (load more fixed on the bottom).

PR #180, opened 2022-02-21
pkp/pkp-lib#7682 Expand DOI filter functionality

ewhanson:


src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -329,16 +329,27 @@ export default {
...
-				// Handle "toggleable" or single select filters
-				newFilters[param] = value;
-			} else {
-				// Handle multi-select filters
-				if (!newFilters[param]) {
-					newFilters[param] = [];
-				}
-				newFilters[param].push(value);
+
+			// Selectable filter options `unpublished`, and `unregistered` all work on a combination of filters.
```
NateWr, line 333: The first line of this comment seems to apply to the block below. I think you could remove it anyway, the code below is well commented.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -329,16 +329,27 @@ export default {
...
 			}
+
+			// The unpublished filter looks for items that are unpublished and have DOIs
+			if (param === 'unpublished') {
+				newFilters['hasDois'] = 1;
+				newFilters = this.addPublishedFilter(newFilters, false);
+				// The unregistered filter looks for items that are published and have DOIs assigned with the STATUS_UNREGISTERED status.
+			} else if (param === 'unregistered') {
+				newFilters['doiStatus'] = pkp.const.DOI_STATUS_UNREGISTERED;
+				newFilters = this.addPublishedFilter(newFilters, true);
```
NateWr, line 349: `this.addPublishedFilter()` creates a fair amount of abstraction and conditional logic for something pretty small. I'd recommend trying to solve this in terms of configuration props that you pass into the `DoiListPanel.vue`. That will keep your Vue.js components small and keep the business logic isolated on the server side, which is usually better. The more generic you can make components the better.

So that would mean adding a `prop` to `DoiListPanel` like:

```
props: {
  publishedStatuses: {
    type: Array,
    required: true,
  },
  publishedStatuses: {
    type: Array,
    required: true,
  },
}
```

Then in the PHP component, you can pass the appropriate values from the server.

src/components/ListPanel/doi/DoiListPanelOJS.vue
```diff
@@ -164,6 +164,54 @@ export default {
...
+		removePublishedFilter(filters) {
+			let newFilters = {...filters};
+
+			if (this.itemType === 'submission') {
+				delete newFilters['status'];
+			} else {
+				delete newFilters['isPublished'];
+			}
+
+			return newFilters;
```
NateWr, line 214: Although less common in JS, you can call "parent" functions so that you have to duplicate less code. Like this:

```js
let newFilters = DoiListPanel.methods.removePublishedFilter.apply(this);
if  (this.itemType === 'issue') {
    delete newFilters['isPublished'];
}
return newFilters;
```

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -365,15 +376,57 @@ export default {
...
-			} else {
-				// Handle multi-select filters
-				newFilters[param] = newFilters[param].filter(v => v !== value);
+
+			// Selectable filter options `unpublished`, and `unregistered` all work on a combination of filters.
+			// `issueIds` is the only filter param that can be combined with any of the other filters, therefore we reset
+			// the primary filter options if we're not removing an `issueIds` filter
+			if (param !== 'issueIds') {
+				delete newFilters['unpublished'];
+				delete newFilters['unregistered'];
```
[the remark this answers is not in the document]
ewhanson: Yeah, the `param: value` combo set in the filters is saved in the active filters, so they need to be cleared manually otherwise they remain. So for unregistered, `unregistered` gets added by default, then we manually add the `doiStatus` and `isPublished`/`status`.

src/components/ListPanel/doi/DoiListPanelOJS.vue
ewhanson: Thanks! Won't be needed anymore as I was able to delete this in favour of passing the filter params in via props as you described above.

PR #192, opened 2022-03-24
pkp/pkp-lib#7743 Add InsertContent and FieldPreparedContent components

NateWr:


src/components/Form/fields/FieldPreparedContent.vue
```diff
@@ -0,0 +1,161 @@
+<template>
+	<field-rich-textarea
+		v-bind="textareaProps"
+		ref="textarea"
+		:init="compiledInit"
+		:value="renderedValue"
+		@change="fieldChanged"
+	>
+		<template slot="footer">
```
ewhanson, line 9: It looks like the `slot="footer"` syntax [has been deprecated](https://v2.vuejs.org/v2/guide/components-slots.html). Would you want to incrementally adopt the new syntax here (`v-slot:footer`) or keep it consistent throughout the code base?

src/components/Form/fields/FieldPreparedContent.vue
```diff
@@ -0,0 +1,161 @@
...
+	},
+	computed: {
+		/**
+		 * Get the props that should be passed down to
+		 * the FieldRichTextarea component
+		 *
+		 * Searches parent components recursively to look
+		 * for accepted props
+		 */
+		textareaProps() {
```
ewhanson, line 88: It seems like this is the case, but I just wanted to make sure I was following this correctly: Is this what's getting and passing along the rest of the required props for `FieldRichTextArea`? I'm thinking specifically about the `plugins` prop which isn't explicitly passed in. I assume this is the case, but wanted to check.

src/components/InsertContent/InsertContent.vue
```diff
@@ -0,0 +1,154 @@
...
+				class="insertContent__item"
+				:id="'insert-content-' + item.key"
+			>
+				<div class="insertContent__item__header">
+					<div class="insertContent__item__value">
+						{{ item.value }}
+					</div>
+					<pkp-button
+						class="insertContent__item_button"
+						aria-describedby="'insert-content-' + item.key"
```
ewhanson, line 21: This is reproduced literally in the HTML as `'insert-content-' + item.key`. I'm guessing it should be rendered as something like `insert-content-contactName`.

src/components/InsertContent/InsertContent.vue
```diff
@@ -0,0 +1,154 @@
...
+
+.insertContent__item__value {
+	font-weight: @bold;
+	padding-right: 1rem;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+}
+
+.insertContent__item_button {
```
ewhanson, line 150: Should this be `insertContent__item__button` to conform to the construction of the other related classes?

src/docs/components/Form/fields/FieldPreparedContent/readme.md
```diff
@@ -0,0 +1,18 @@
+## Props
+
+| Key | Description |
+| --- | --- |
+| `...` | Supports all props in [FieldRichTextarea](#/component/Form/fields/FieldRichTextarea). |
+| `preparedContent` | An optional object containing preset information. When preset information exists, a button will appear in the toolbar. See the [Prepared Content](#/component/Form/fields/FieldRichTextarea/ExamplePreparedContent) example. |
```
ewhanson, line 6: The `Prepared Content` example goes to the removed example. Should it now point to `#/component/Form/fields/FieldPreparedContent`?

src/components/Form/fields/FieldPreparedContent.vue
NateWr: Thanks, I didn't know this! I've got it filed at https://github.com/pkp/pkp-lib/issues/7801. In the meantime, I'm happy to keep things consistent.

src/components/Form/fields/FieldPreparedContent.vue
NateWr: Yep, exactly that. This is an unusual setup. It's closer to a parent-child relationship that you see in PHP, and that's not common in JS or Vue.js, so I'm doing some acrobatics to make it work.

Since `FieldPreparedContent` extends `FieldRichTextarea`, it inherits all of the props, data, methods, etc, like a normal object-oriented parent-child relationship. However, `FieldPreparedContent` _uses_ `FieldRichTextarea` in its template. This is a Vue.js-style parent-child relationship, but totally different from the object-oriented one. 

So this code gathers up all of the props for `FieldPreparedContent` that are required by `FieldRichTextarea`, so that it can pass them down as props. The trick is that, in Vue.js, the `extends` relationship is only applied when the component is instantiated. The JS object `FieldPreparedContent` _doesn't_ contain the props of `FieldRichTextarea` until it is instantiated as a component.

This means that I can't do `FieldPrepardContent.props.plugins`. That property doesn't exist because `FieldPreparedContent.props` only exists on the component's configuration object (ie - what we code up), not the actual component that is instantiated. I have to do this extra bit of work to look in the parent components to determine all of the accepted props.

(If you've miraculously understood this and thought two steps ahead, you might be wondering why I can't just gather all props of `FieldPreparedContent` and pass them to `<field-rich-textarea>`. That technically works, but it causes extra attributes in the HTML code. When `<field-rich-textarea>` receives a prop that it doesn't recognize, it adds it as an attribute, like `<div class="fieldRichTextarea" prepared-content="[Object]">`.)

Not sure if that made sense! This is really getting into the deep underbelly of Vue. :laughing:

PR #197, opened 2022-04-19
6782: institutions

bozana:
This PR is part of usage stats improvements, adding institution list panel

### DIFF
--- a/src/components/Container/SettingsPage.vue
+++ b/src/components/Container/SettingsPage.vue
@@ -1,6 +1,7 @@
 <script type="text/javascript">
 import Page from './Page.vue';
 import AnnouncementsListPanel from '@/components/ListPanel/announcements/AnnouncementsListPanel.vue';
+import InstitutionsListPanel from '@/components/ListPanel/institutions/InstitutionsListPanel.vue';
 import EmailTemplatesListPanel from '@/components/ListPanel/emailTemplates/EmailTemplatesListPanel.vue';
 import PkpForm from '@/components/Form/Form.vue';
 import ThemeForm from '@/components/Form/context/ThemeForm.vue';
@@ -11,6 +12,7 @@ export default {
 	extends: Page,
 	components: {
 		AnnouncementsListPanel,
+		InstitutionsListPanel,
 		EmailTemplatesListPanel,
 		PkpForm,
 		ThemeForm,
--- a/src/components/Container/SettingsPage.vue
+++ b/src/components/Container/SettingsPage.vue
@@ -21,6 +21,7 @@ export default {
 	data() {
 		return {
 			announcementsNavLink: {},
+			institutionsNavLink: {},
 			paymentsNavLink: null
 		};
 	},
@@ -42,7 +43,31 @@ export default {
 					let menu = {};
 					Object.keys(this.menu).forEach(key => {
 						if (key === 'settings') {
+							if (!this.menu['institutions']) {
+								menu.institutions = this.institutionsNavLink;
+							}
 							menu.payments = this.paymentsNavLink;
+						}
+						menu[key] = this.menu[key];
+					});
+					this.menu = menu;
+				}
+			}
+
+			// Add or remove institutions nav link
+			if (formId === pkp.const.FORM_CONTEXT_STATISTICS) {
+				if (!context.enableInstitutionUsageStats && !this.menu['payments'] && !!this.menu['institutions']) {
+					let menu = {...this.menu};
+					delete menu.institutions;
+					this.menu = menu;
+				} else if (
+					context.enableInstitutionUsageStats &&
+					!Object.keys(this.menu).includes('institutions')
+				) {
+					let menu = {};
+					Object.keys(this.menu).forEach(key => {
+						if (key === 'settings' || key === 'payments') {
+							menu.institutions = this.institutionsNavLink;
 						}
 						menu[key] = this.menu[key];
 					});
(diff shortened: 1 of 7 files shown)


src/components/Container/SettingsPage.vue
```diff
@@ -40,6 +43,9 @@ export default {
 					let menu = {};
 					Object.keys(this.menu).forEach(key => {
 						if (key === 'settings') {
+							if (!this.menu['institutions']) {
+								menu.institutions = this.institutionsNavLink;
+							}
```
NateWr, line 48: No need for the `if`. If the link already exists it will just replace it.

src/components/Container/SettingsPage.vue
```diff
@@ -60,7 +87,7 @@ export default {
 				) {
 					let menu = {};
 					Object.keys(this.menu).forEach(key => {
-						if (key === 'settings' || key === 'payments') {
+						if (key === 'settings' || key === 'payments' || key === 'institutions') {
```
NateWr, line 90: I think maybe there are some gaps here, like when payments are disabled the institutions might not be removed (line 37).

I'd suggest simplifying this logic and just checking the `context` object once for each nav link, rather than trying to tie the items up together. Something like this:

```js
if (!context.paymentsEnabled && !!this.menu.payments) {
	let menu = {...this.menu};
	delete menu.payments;
	this.menu = menu;
} else if (!this.menu.payments) {
	// add payments
}

if (!context.paymentsEnabled && !context.enableInstitutionUsageStats && !!this.menu.institutions) {
	let menu = {...this.menu};
	delete menu.institutions;
	this.menu = menu;
} else if (!this.menu.institutions) {
	// add institutions
}

if (!context.enableAnnouncements && !!this.menu.announcements) {
	let menu = {...this.menu};
	delete menu.announcements;
	this.menu = menu;
} else if (!this.menu.announcements) {
	// add institutions
}
```

src/components/ListPanel/institutions/InstitutionsListPanel.vue
```diff
@@ -0,0 +1,305 @@
...
+</script>
+
+<style lang="less">
+@import '../../../styles/_import';
+
+.institutionsListPanel {
+	.listPanel__item {
+		padding-top: 0.25rem;
+		padding-bottom: 0.25rem;
+	}
```
NateWr, line 299: Shouldn't need the padding here.

src/components/ListPanel/institutions/InstitutionsListPanel.vue
```diff
@@ -0,0 +1,305 @@
...
+				<template v-slot:itemTitle="{item}">
+					{{ localize(item.name) }}
+				</template>
+				<template v-slot:itemActions="{item}">
+					<pkp-button @click="openEditModal(item.id)">
+						{{ __('common.edit') }}
+					</pkp-button>
+					<pkp-button :isWarnable="true" @click="openDeleteModal(item.id)">
+						{{ __('common.delete') }}
+					</pkp-button>
```
NateWr, line 27: For accessibility, we need to link the button labels to the name of the item they edit/delete. The way I've been doing that lately is to use the `aria-describedby` attribute. First, put an id on the item name:

```html
<template v-slot:itemTitle="{item}">
    <span :id="'institution-' + item.id">
        {{ localize(item.name) }}
    </span>
</template>
```

Then reference this ID in the button:

```html
<pkp-button :aria-describedby="'institution-' + item.id" @click="...">
    {{ __('common.edit') }}
</pkp-button>
```

src/components/ListPanel/institutions/InstitutionsListPanel.vue
```diff
@@ -0,0 +1,305 @@
+<template>
```
NateWr, line 1: We'll need to add this component to the UI Library. You can read about how to do that at https://docs.pkp.sfu.ca/dev/ui-library/dev/#/pages/contributing. Then probably the best thing to do is to look at a similar component, like `AnnouncementListPanel`, and use that as an example to create the component examples and add them to the library.

NateWr (changes requested):

src/components/Container/SettingsPage.vue
bozana: I am not 100% sure that I fully understand: This code is used when a form is saved. When payment form is saved, it will not contain enableInstitutionUsageStats, and the vice versa. So there is no possibility to check enableInstitutionUsageStats in order to know if it should be deleted when saving the payment form.
I believe we decided to leave it so -- with the next URL reload it will be correctly removed...
Or, maybe I do not understand exactly how you mean it?

NateWr: Whenever any of the context forms are saved, the full context object is returned. So it should be possible to check any context setting, even if the form didn't modify that setting.

bozana: Hmmm... I do get the full context object when e.g. saving the statistics or announcements form, but when I save the payments form I do not get that... only:
```
Object { paymentsEnabled: true, currency: "", paymentPluginName: "", testMode: "false", accountName: "", clientId: "", secret: "", manualInstructions: "" }
accountName: ""
clientId: ""
currency: ""
manualInstructions: ""
paymentPluginName: ""
paymentsEnabled: true
secret: ""
testMode: "false"
<prototype>: Object { … }
```
I can investigate...
Maybe you would know immediately?
Thanks a lot!

NateWr: Oh I see. :/ Ok, then leave it alone. Sorry for the confusion!

bozana:
Hi @NateWr, I've addressed your comments in this PR.
The check for navigation link, that uses the context object, stayed the same because the context object seems to be different in payments form...
For some reason also the style
```
.institutionsListPanel {
	.listPanel__itemTitle {
		font-weight: @normal;
	}
}
```
is not applied, when viewing the institutions list panel in OJS, so that the institution name is bigger (font-weight: 700).
Thanks a lot!!!

NateWr:
I just checked and the CSS is working for me. Maybe you ran into a funny cache issue. The hot-reloading in the UI Library can be funny sometimes.

bozana:
closing the PR -- it was for the first review round

PR #209, opened 2022-06-29
6782: institutions

bozana:
This PR is part of usage stats improvements, adding institution list panel

### DIFF
--- a/src/components/ListPanel/institutions/InstitutionsListPanel.vue
+++ b/src/components/ListPanel/institutions/InstitutionsListPanel.vue
@@ -16,13 +16,22 @@
 					</template>
 				</pkp-header>
 				<template v-slot:itemTitle="{item}">
-					{{ localize(item.name) }}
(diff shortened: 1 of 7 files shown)


src/components/ListPanel/institutions/InstitutionsListPanel.vue
```diff
@@ -0,0 +1,303 @@
+<template>
```
asmecher, line 1: This is more than likely something wrong with my checkout, but just in case: the institutions modal is showing up without padding for me. (The text and fields almost touch the sides of the modal.)
image

NateWr: If I recall, I think I suggested to @bozana to merge it as is and I can fix it later. It's a CSS problem with forms that appear inside of modals that I need to resolve across the ui library.

bozana:
closing this pr -- it was for the second review round

PR #214, opened 2022-07-24
pkp/pkp-lib#8027 OPS preprints should support DOI versioning

ewhanson:


src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -74,19 +74,38 @@
 							class="pkpFormField__input pkpFormField--text__input"
 							:id="row.uid"
 							type="text"
-							:disabled="!(isEditingDois && !isSaving)"
+							:readonly="!(isEditingDois && !isSaving)"
```
ewhanson, line 77: @NateWr, this is one instances where we changed `disabled` to `readonly`. Would it be best to update the styles for this globally somewhere, or add something for this component specifically?

NateWr: Yeah, for now, let's just make `readonly` and `disabled` show the same styles. In `FieldBase.vue`, find the styles for `.pkpFormField__input`, and change:

```css
&:disabled {
```

to:

```css
&:disabled,
&:read-only {
```

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -64,7 +64,7 @@
 			v-if="isExpanded"
 			class="listPanel__itemExpanded listPanel__itemExpanded--doi"
 		>
-			<pkp-table :columns="doiListColumns" :rows="item.doiObjects">
+			<pkp-table :columns="doiListColumns" :rows="item.doiObjects.filter(doiObject => doiObject.isCurrentVersion)">
```
NateWr, line 67: Maybe use a computed property for this:

```js
currentVersionDoiObjects() {
  return this.item.doiObjects.filter(...);
}
```

```html
<pkp-table :columns="doiListColumns" :rows="currentVersionDoiObjects">
```

src/components/ListPanel/doi/DoiListPanelOPS.vue
```diff
@@ -14,59 +14,28 @@ export default {
...
+					publication.id === this.getCurrentPublication(originalItem).id;
 
-				let updateWithNewDoiEndpoint = `${this.doiApiUrl}/publications/${originalItem.currentPublicationId}`;
-				updateWithNewDoiEndpoint = updateWithNewDoiEndpoint.replace(
-					/dois/g,
-					'_dois'
-				);
+				// Submissions
+				if (this.enabledDoiTypes.includes('publication')) {
+					let doiObject = publication.doiObject;
```
NateWr, line 23: It looks like you never modify `doiObject` so you can make it `const doiObject = ...` here and below.

src/components/ListPanel/doi/DoiListPanelOPS.vue
```diff
@@ -83,8 +52,46 @@ export default {
...
+								doiObject === null
+									? null
+									: doiObject[
+											this.registrationAgencyInfo['registeredMessageKey']
+									  ],
+							updateWithNewDoiEndpoint: updateWithNewDoiEndpoint
+						});
+					});
+				}
+			});
```
NateWr, line 94: You'll save yourself some pain if you create a helper method in the parent `DOIListPanel.vue` that helps you build these mapped objects. Something like:

```js
mapDoiObject(doiObject, props) {
	return {
		doiId: doiObject === null ? null : doiObject.id,
		identifier: doiObject === null ? '' : doiObject.doi,
		depositStatus: ...,
		...props
	};
},
```

Then you can use it in each app's list panel like this:

```js
if (this.enabledDoiTypes.includes('publication')) {
    const doiObject = publication.doiObject;

    const updateWithNewDoiEndpoint = `${this.doiApiUrl}/publications/${originalItem.currentPublicationId}`;
    updateWithNewDoiEndpoint = updateWithNewDoiEndpoint.replace(
        /dois/g,
        '_dois'
    );

    newMappedItem.doiObjects.push(
        this.mapDoiObject(doiObject, {
            id: publication.id,
            uid: ...,
            identifier: ...,
        })
    );
    ...
```

src/components/ListPanel/doi/DoiListItem.vue
ewhanson: Hey @NateWr, I tried this, and it's applying the `.pkpFormField__input:read-only` styles to the `<select>` tags in the `FieldSelect` components, and this is happening throughout the application. I'm not sure if it's just considering the `<select>` element to be read only for some reason. I'll leave this added in the PR for now, but can remove it if it's more trouble than it's worth trying to add it to `FieldBase` rather than specific styling in `DoiListItem`.

NateWr: That's interesting... I think the `<select>` fields don't support `readonly`, so I wonder if the CSS parser is evaluating it as `true` for some reason. I'll take a look in my code review.

bozana:
This all looks good to me :-) :+1:

PR #216, opened 2022-10-05
pkp/pkp-lib#7318 publication stats export

bozana:

### DIFF
--- a/src/components/Container/StatsPublicationsPage.vue
+++ b/src/components/Container/StatsPublicationsPage.vue
@@ -38,68 +38,6 @@ export default {
 	},
 	computed: {
 		/**
-		 * Get description active filters titles to display on the export modal
-		 *
-		 * @return string
-		 */
-		getFiltersTitlesDescription() {
-			let returnStr = [];
-			this.filters.forEach(filterSet => {
-				let filterTitles = [];
-				filterSet.filters.forEach(element => {
-					if (
-						element.param in this.activeFilters &&
-						this.activeFilters[element.param].includes(element.value)
-					) {
-						filterTitles.push(element.title);
-					}
-				});
-				let str = this.replaceLocaleParams(this.inAllFiltersLabel, {
-					filter: filterSet['heading'].toLowerCase()
-				});
-				if (filterTitles.length != 0) {
-					str = this.replaceLocaleParams(this.inFiltersLabel, {
-						filter: filterTitles.join(', ')
-					});
-				}
-				returnStr.push(str);
-			});
-			if (!returnStr.length) {
-				return '';
-			}
-			return returnStr.join(' ' + this.commonAndLabel + ' ');
-		},
-
-		/**
-		 * Get date range description to display on the expot modal
-		 *
-		 * @return string
-		 */
-		getDateRangeDescription() {
-			if (!this.dateStart && !this.dateEnd) {
-				return this.allDatesLabel;
-			}
-			return this.replaceLocaleParams(this.betweenDatesLabel, {
-				startDate: this.dateStart,
-				endDate: this.dateEnd
-			});
-		},
-
-		/**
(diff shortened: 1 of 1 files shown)


src/components/Container/StatsPublicationsPage.vue
```diff
@@ -138,6 +147,133 @@ export default {
...
+				(type ? type : 'submissions') +
+				'_' +
+				this.getDateRangeDescription()
+					.replaceAll(' ', '')
+					.replace('to', '_') +
+				'_' +
+				filterTitles.join('_') +
+				'_' +
+				this.searchPhrase +
+				'.csv';
```
NateWr, line 250: A small recommendation: instead of concatenating a bunch of strings together, try joining an array:

```js
let downloadFileName = [
		'stats_',
		new Date()
			.toISOString()
			.substr(0, 19)
			.replaceAll(':', '-'),
		(type ? type : 'submissions'),
		this.getDateRangeDescription()
			.replaceAll(' ', '')
			.replace('to', '_'),
		filterTitles.join('_'),
		this.searchPhrase
	]
	.join('_')
	+ '.csv';
```

NateWr: This will also help fix issues where nothing is selected, like `this.searchPhrase`, which causes `null` in the filename:

```
stats_2022-10-11T10-15-32_submissions_Alldates_AllSections_null.csv
```

The following will remove empty values before joining the array:

```js
let downloadFileName = [
		'stats',
		new Date()
			.toISOString()
			.substr(0, 19)
			.replaceAll(':', '-'),
		(type ? type : 'submissions'),
		this.getDateRangeDescription()
			.replaceAll(' ', '')
			.replace('to', '_'),
		filterTitles.join('_'),
		this.searchPhrase ?? '',
	]
	.filter(i => i) // removes empty values. it is the same as function(i) => { return $i ? $i : false}
	.join('_')
	+ '.csv';
```

src/components/Container/StatsPublicationsPage.vue
```diff
@@ -138,6 +147,133 @@ export default {
...
+						.replaceAll(' ', '')
+						.replaceAll(',', '_')
+				);
+			});
+			let downloadFileName =
+				`stats_` +
+				new Date()
+					.toISOString()
+					.substr(0, 19)
+					.replaceAll(':', '-') +
```
NateWr, line 239: Let's remove the current date from the filename for now. This will be available in the file data (when the file was saved) and it makes it hard to understand the other dates in the filename.

src/components/Container/StatsPublicationsPage.vue
```diff
@@ -138,6 +147,133 @@ export default {
...
+					this.activeFilters[element.param].includes(element.value)
+				) {
+					filterTitles.push(element.title);
+				}
+			});
+			let description = this.replaceLocaleParams(this.allFiltersLabel, {
+				filter: filterSet.heading
+			});
+			if (filterTitles.length != 0) {
+				description = filterTitles.join(', ');
```
NateWr, line 170: Use `.join(this.__('common.commaListSeparator'))` here. That locale string is available globally in the frontend.

src/components/Container/StatsPublicationsPage.vue
```diff
@@ -138,6 +147,133 @@ export default {
...
+			$.ajax({
+				url: this.apiUrl + (type ? '/' + type : ''),
+				type: 'GET',
+				context: this,
+				headers: {
+					Accept: 'text/csv; charset=utf-8',
+					'Content-Type': 'text/csv;·charset_utf-8'
+				},
+				data: this.getReportParams(),
+				_uuid: this.latestItemsGetRequest,
```
NateWr, line 261: This can be removed.

PR #217, opened 2022-10-11
pkp/pkp-lib#7318 stats issues page

bozana:
This is the vue page for the Statistics > Issues page


src/components/Container/StatsIssuesPage.vue
```diff
@@ -0,0 +1,449 @@
+<script>
+import StatsPage from '@/components/Container/StatsPage.vue';
+import ActionPanel from '../ActionPanel/ActionPanel.vue';
+import LineChart from '@/components/Chart/LineChart.vue';
+import Search from '@/components/Search/Search.vue';
+import debounce from 'debounce';
+
+export default {
+	name: 'StatsIssuesPage',
+	extends: StatsPage,
```
NateWr, line 10: I think we should extend `StatsPublicationsPage` and just override the things that are different. Otherwise there is a lot of duplicated code between this and the publications page component.

NateWr (changes requested):

src/components/Container/StatsIssuesPage.vue
```diff
@@ -0,0 +1,55 @@
...
+					link.download = downloadFileName;
+					link.click();
+				},
+				complete(r) {
+					this.isDownloadingReport = false;
+					this.$modal.hide('export');
+				}
+			});
+		}
+	}
```
NateWr, line 53: Rather than duplicating the full `downloadReport()` method, which includes the ajax call and other things, just split out the bit of code that you need to modify into its own method. I _think_ that's just the download file name, so in `StatsPublicationsPage` you can do:

```js
let downloadFileName = this.getReportFilenameParts()
	.filter(i => i)
	.join('_') + '.csv';
```

Then you only have to override `getReportFilenameParts()` in the stats page. (you'll probably do something similiar with the context stats page).

src/components/Container/StatsPublicationsPage.vue
```diff
@@ -231,16 +233,25 @@ export default {
 						.replaceAll(',', '_')
 				);
 			});
+			return [
+				'stats',
+				type ? type : 'submissions',
+				this.getDateRangeDescription()
+					.replaceAll(' ', '')
+					.replace('to', '_'),
```
NateWr, line 241: It just occurred to me that this probably won't work here or in `StatsIssuesPage.vue`. The `to` is part of a string that will be translated so won't appear in most languages.

bozana: Oh, yes :-DDD Good catch!!!
I will then change that...
Also, I am wondering if having issue identifications in the file name could lead to the long and thus unreadable file names -- e.g. if the journal uses/has only issue title...
Thus, what do you think about having the file name that only contains e.g. stats_submissions_currentDateTime, and adding the parameters in the CSV file in the same way we list them on the download modal, e.g.:

```
"Date Range","All dates"
Sections,AllSections
Issues,AllIssues
"Search Phrase",anti

ID,Title,Total,"Abstract Views","File Views",PDF,HTML,Other
17,"The Antimicrobial, heavy metal resistance and plasmid profile of coliforms isolated from nosocomial infections in a hospital in Isfahan, Iran",14,14,0,0,0,0
```

bozana: Only timeline report considers only either views or downloads, the other report consider both. Thus I will then add timeline type and timeline interval to that report. The others will have only date range, filters (if exist) and search phrase.

PR #222, opened 2022-10-22
fix getFilterDescription()

bozana:


NateWr (approved):
This looks good but the commit message is missing the issue identification: `pkp/pkp-lib#...`

PR #225, opened 2022-10-25
[UI-LIBRARY][main] #7397 Display fixes for submissions with no contributors

defstat:


src/components/ListPanel/submissions/SubmissionsListItem.vue
```diff
@@ -877,10 +877,10 @@ export default {
 
 .listPanel__item--submission__id {
 	position: absolute;
-	top: 0;
+	top: 50%;
 	left: 0;
 	font-size: @font-tiny;
-	line-height: 22px; // Match baseline of title
+	transform: translateY(-50%);
```
NateWr, line 883: What's the reason for the CSS changes here and in `CatalogListItem.vue`?

src/docs/components/WorkflowPage/previews/PreviewWorkflowPage.vue
```diff
@@ -18,10 +18,10 @@
 				</badge>
 				<span class="pkpWorkflow__identificationId">{{ submission.id }}</span>
 				<span class="pkpWorkflow__identificationDivider">/</span>
-				<span class="pkpWorkflow__identificationAuthor">
+				<span v-if="!!currentPublication.authorsStringShort" class="pkpWorkflow__identificationAuthor">
 					{{ currentPublication.authorsStringShort }}
 				</span>
-				<span class="pkpWorkflow__identificationDivider">/</span>
+				<span v-if="!!currentPublication.authorsStringShort" class="pkpWorkflow__identificationDivider">/</span>
```
NateWr, line 24: The `!!` is not necessary here or above. Vue's templating language will handle `v-if="currentPublication.authorsStringShort"` just the same.

NateWr (changes requested):

src/components/ListPanel/submissions/SubmissionsListItem.vue
defstat: The original code's display result for a submission in `SubmissionListPanel` (and in the `CatalogListPanel`) is as follows:
image

The `id` seemed a little bit "off" in my opinion, so I am proposing this change, that renders the id vertically in the middle of whatever height the container will have, resulting something like this for submissions with no contributors
image

and something like this for submissions with contributors
image

src/docs/components/WorkflowPage/previews/PreviewWorkflowPage.vue
defstat: Changed

src/components/ListPanel/submissions/SubmissionsListItem.vue
NateWr: Ah, I see. The way it was designed was intended to ensure that the baseline matched with the first line. So the bottom of the ID text should match the bottom of the text on the right, rather than be vertically centred.

I think instead of changing the styles, you should be able to fix it when no author name exists by adding a `v-if` to `SubmissionsListItem.vue` that removes the `.listPanel__itemTitle` when no title exists. That should move the article title up and the baseline should be kept.

defstat: That was my first instinct also (to remove the `.listPanel__itemTitle`) but, because of the `listPanel__item--submission__id`'s `top: 0;` css directive it does not render next to the subtitle.

image

NateWr: Filed this at https://github.com/pkp/pkp-lib/issues/8370.

defstat: Sure, removed

PR #226, opened 2022-10-29
pkp/pkp-lib#7318 Stats Context Page

bozana:


src/components/Container/StatsContextPage.vue
```diff
@@ -0,0 +1,52 @@
...
+export default {
+	name: 'StatsContextPage',
+	extends: StatsPublicationsPage,
+	computed: {
+		/**
+		 * The params to send with each GET request
+		 *
+		 * @return Object
+		 */
+		getParams() {
```
ewhanson, line 13: This should be a method now instead of a computed property (It's defined as a method in `StatsPublicationsPage`).

bozana: Oh, yes, good catch! Thanks!!! :+1:

PR #232, opened 2022-11-22
Bugfix/main/8442 fix date filter

jonasraoni:


src/components/Container/StatsEditorialPage.vue
```diff
@@ -86,99 +86,91 @@ export default {
...
 			// are back.
-			let dateRangeFinished = false;
-			let totalsFinished = false;
-			let averagesFinished = false;
 			let dateRangeResponse = [];
 			let totalsResponse = [];
 			let averagesResponse = {};
 
+			let NetworkError = function(response) {
+				this.response = response;
```
NateWr, line 94: Rather than sticking this into the component `data`, can we use a scoped variable so it doesn't effect anything outside of this function?

```js
let response;
let NetworkError = function(r) {
	response = r;
}
```

That way it won't clutter up the component data.

src/components/DateRange/DateRange.vue
```diff
@@ -149,6 +149,10 @@ export default {
 			type: String,
 			required: true
 		},
+		invalidEndDateMaxLabel: {
+			type: String,
+			required: true
+		},
```
NateWr, line 155: It looks like `invalidEndDateMaxLabel` may not be used?

NateWr (changes requested):
Nice use of promises. I hope to refactor away from `$.ajax()` to `fetch()` soon. Just a couple of small comments.

src/components/DateRange/DateRange.vue
jonasraoni: Ah, this is already on the `main` branch:
https://github.com/pkp/ui-library/blob/a47e575b21e53913c1ef1cf07e0445c9c5668a2d/src/components/DateRange/DateRange.vue#L263

An error is generated when the code reaches that point, after the fix it will work fine.

src/components/Container/StatsEditorialPage.vue
jonasraoni: Given the function isn't an arrow function, `this` will be "self-contained", as I'm instantiating it with `new NetworkError` later in the code.
p.s.: The errored response isn't used, but I didn't want to lose information :)

PR #234, opened 2022-12-01
pkp/pkp-lib#8336 Improve editorial decisions UI

NateWr:


src/components/Composer/Composer.vue
```diff
@@ -508,6 +521,24 @@ export default {
 				this.emitChange({bcc: newVal});
 			}
 		},
+		bodyInit() {
+			if (!this.attachers.length) {
+				return {};
+			}
+			let self = this;
+			return {
+				setup: function(editor) {
```
jonasraoni, line 530: If you want to get rid of the `self`, I think it's ok to transform this into an arrow function, then replace `self` by `this`.

src/components/Composer/Composer.vue
```diff
@@ -674,6 +705,20 @@ export default {
...
 		},
 
+		/**
+		 * Get a plain text snippet of an email template's body
+		 */
+		getBodySnippet(str) {
+			const length = 70;
+			let span = document.createElement('span');
+			span.innerHTML = str;
+			const snippet = (span.textContent || span.innerText).trim();
```
jonasraoni, line 715: I think it's ok to leave just `textContent` here. As `innerText` is a kind of limited `textContent`, which will skip non-visible elements 🤔

src/components/Composer/Composer.vue
```diff
@@ -674,6 +705,20 @@ export default {
...
+			if (snippet.length < length - 3) {
+				return snippet;
+			}
+			return snippet.substring(0, 80).trim() + '...';
```
jonasraoni, line 719: Here it shouldn't be `length - 3` instead of `80`?

src/components/ListPanel/submissionFiles/SelectSubmissionFileListItem.vue
```diff
@@ -84,6 +102,19 @@ export default {
...
 		};
 	},
+	computed: {
+		uploadedDetails() {
+			if (this.uploadedBy && this.createdAt) {
+				return this.__('common.uploadedByAndWhen', {
+					name: this.uploadedBy,
+					date: this.createdAt.substring(0, 10)
+				});
+			} else if (this.uploadedBy) {
```
jonasraoni, line 112: You might drop the `else` if you like early returns :)

src/components/Steps/Steps.vue
```diff
@@ -95,6 +107,13 @@ export default {
 		};
 	},
 	computed: {
+		/**
+		 * The steps that have been completed
+		 */
+		completedSteps() {
+			return this.startedSteps.slice(0, this.startedSteps.length - 1);
```
jonasraoni, line 114: Also not important, but you might use `.slice(0, -1)`

jonasraoni (approved):

src/components/Composer/Composer.vue
NateWr: I think you're right, but I vaguely recall running into a problem with that. TinyMCE is funny in places. I'm going to leave this as-is just to save the effort of testing againt, but it's a good idea.

PR #247, opened 2023-01-06
pkp/pkp-lib#8485 Some strings not translated in DOI Management UI

ewhanson:


bozana (commented):
Just one question :-)

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -246,14 +246,14 @@ export default {
...
-					label: 'Type',
+					label: this.__('common.type'),
 					value(row) {
 						return row.displayType;
 					}
 				},
 				{
 					name: 'doi',
-					label: 'DOI',
+					label: this.__('manager.dois.title'),
```
bozana, line 256: I suppose those manager.dois locale keys are somewhere provided so that one can access them this way, similar to e.g. "article.article"?

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -86,7 +86,7 @@
 					:is-disabled="isDeposited || isSaving"
 					@click="isEditingDois ? saveDois() : editDois()"
 				>
-					{{ isEditingDois ? 'Save changes' : 'Edit DOI(s)' }}
+					{{ isEditingDois ? __('common.edit') : __('common.save') }}
```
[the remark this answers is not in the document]
ewhanson: Oops, thanks!

src/components/ListPanel/doi/DoiListItem.vue
ewhanson: Yes, the locale is in `manager.po` in pkp-lib here: https://github.com/pkp/pkp-lib/blob/ef79d952408a9c91fa0b38cfbdb407df69429704/locale/en_US/manager.po#L352-L353. It is set to the Vue component in `PKPDoiListPanel` here: https://github.com/ewhanson/pkp-lib/blob/397cc76629a874cdba4d8b390a9c4104e5485341/classes/components/listPanels/PKPDoiListPanel.php#L178.

PR #248, opened 2023-01-09
pkp/pkp-lib#8306 Failed jobs UI

touhidurabir:
This PR aims to provide solution for pkp/pkp-lib#8306

### DIFF
--- a/src/components/Container/FailedJobsPage.vue
+++ b/src/components/Container/FailedJobsPage.vue
@@ -2,6 +2,7 @@
 import Page from './Page.vue';
 import Pagination from '@/components/Pagination/Pagination.vue';
 import PkpTable from '@/components/Table/Table.vue';
+import TableCell from '@/components/Table/TableCell.vue';
 import ajaxError from '@/mixins/ajaxError';
 
 export default {
@@ -9,6 +10,7 @@ export default {
 	extends: Page,
 	components: {
 		PkpTable,
+		TableCell,
 		Pagination
 	},
 	data() {
@@ -16,7 +18,6 @@ export default {
 			columns: [],
 			rows: [],
 			label: '',
-			description: '',
 			total: 0,
 			currentPage: 1,
 			lastPage: 1,
@@ -25,24 +26,23 @@ export default {
 		};
 	},
 	mixins: [ajaxError],
+	computed: {
+		description() {
+			return this.__('admin.jobs.failed.totalCount', {total: this.total});
+		}
+	},
 	methods: {
-		handlePagination: function(page) {
+		handlePagination(page) {
 			this.isLoadingItems = true;
-			const paged = new URL(window.location.href);
-			paged.searchParams.set('page', page);
-			window.location = paged;
+			this.loadList(page);
 		},
-		refreshComponent: function(data) {
+		updateJobList(data) {
 			this.rows = this.rows.filter(row => row.id !== data.id);
 			this.total = this.total - 1;
-			this.description = this.__('admin.jobs.failed.totalCount', {
-				total: this.total
-			});
-			this.$forceUpdate();
 		},
-		redispatch: function(data) {
+		redispatch(data) {
 			$.ajax({
-				url: this.apiUrl + '/redispatch/' + data.id,
+				url: data._hrefs._redispatch,
 				type: 'POST',
 				headers: {
 					'X-Csrf-Token': pkp.currentUser.csrfToken
@@ -50,13 +50,13 @@ export default {
 				error: this.ajaxErrorCallback,
 				success: response => {
 					pkp.eventBus.$emit('notify', response.message, 'success');
-					this.refreshComponent(data);
+					this.updateJobList(data);
 				}
 			});
 		},
-		remove: function(data) {
+		remove(data) {
 			$.ajax({
-				url: this.apiUrl + '/failed/delete/' + data.id,
+				url: data._hrefs._delete,
 				type: 'POST',
 				headers: {
 					'X-Csrf-Token': pkp.currentUser.csrfToken,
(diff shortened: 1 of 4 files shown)


src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,83 @@
...
+			total: 0,
+			currentPage: 1,
+			lastPage: 1,
+			isLoadingItems: false,
+			apiUrl: null
+		};
+	},
+	mixins: [ajaxError],
+	methods: {
+		handlePagination: function(page) {
```
NateWr, line 29: For each of these methods, we prefer the shortened syntax:

```js
handlePagination(page) {
```

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,83 @@
...
+			this.isLoadingItems = true;
+			const paged = new URL(window.location.href);
+			paged.searchParams.set('page', page);
+			window.location = paged;
+		},
+		refreshComponent: function(data) {
```
NateWr, line 35: Should this method be renamed `removeJob(job)`?

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,83 @@
...
+			this.rows = this.rows.filter(row => row.id !== data.id);
+			this.total = this.total - 1;
+			this.description = this.__('admin.jobs.failed.totalCount', {
+				total: this.total
+			});
+			this.$forceUpdate();
```
NateWr, line 41: You should try to avoid this whenever possible, because it will force the entire component to redraw. Vue.js can intelligently redraw only the HTML elements that need to be redrawn, which is much better for performance.

A redraw should be happening for you without requiring this. `this.rows = ...` should be triggering it. If not, we can look into that.

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,83 @@
...
+				type: 'POST',
+				headers: {
+					'X-Csrf-Token': pkp.currentUser.csrfToken,
+					'X-Http-Method-Override': 'DELETE'
+				},
+				error: this.ajaxErrorCallback,
+				success: response => {
+					pkp.eventBus.$emit('notify', response.message, 'success');
+					this.refreshComponent(data);
+				}
```
NateWr, line 69: When designing the API, I encourage you to rely less on success messages and more on success statuses and data. For example, this action is similar in some respects to our `DELETE` methods. When we `DELETE` an item, we return the item we deleted with a `200` response. This says "successfully deleted" and gives the handling code information about what was just deleted.

Calling code can then use the response data, rather than relying on request data:

```js
success: (response) => {
    const successMessage = this.__('...', {name: response.name});
    pkp.eventBus.$emit('notify', successMessage, 'success');
    this.rows = this.rows.filter(row => row.id !== response.id);
}
```

The difference is not very significant for us. But the result is an API response that is more useful in different scenarios, for example if you only had access to the failed job ID, not the full object.

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,83 @@
...
+			const paged = new URL(window.location.href);
+			paged.searchParams.set('page', page);
+			window.location = paged;
+		},
+		refreshComponent: function(data) {
+			this.rows = this.rows.filter(row => row.id !== data.id);
+			this.total = this.total - 1;
+			this.description = this.__('admin.jobs.failed.totalCount', {
+				total: this.total
+			});
```
NateWr, line 40: In general, it's best to use computed properties to derive information like the `description`. When you do this, Vue.js will automatically recompile it when `total` changes. That way, no matter when or where that value changes, `description` is always in-sync:

```js
computed: {
    description() {
        return this.__('...', {total: this.total});
    }
}
```

src/components/Table/Table.vue
```diff
@@ -60,6 +60,8 @@
 						:tabindex="!rowIndex && !columnIndex ? 0 : -1"
 					></table-cell>
 				</slot>
+
+				<slot name="actions" :row="row" />
```
NateWr, line 64: Rather than adding another slot to this component, I'd prefer if any actions were added to a column. You can find an example of adding a button to a table column here: https://docs.pkp.sfu.ca/dev/ui-library/dev/#/component/Table/with-custom-column

src/components/Container/FailedJobsPage.vue
touhidurabir: done.

src/components/Container/FailedJobsPage.vue
touhidurabir: done.

src/components/Container/FailedJobsPage.vue
touhidurabir: fixed.

src/components/Container/FailedJobsPage.vue
touhidurabir: Done but it seems it can not find the translation in JS end. for example , the key `admin.jobs.failed.totalCount` present in the `lib/pkp/locale/en_US/admin.po` file , am I missing something ?

src/components/Table/Table.vue
touhidurabir: done .

src/components/Container/FailedJobsPage.vue
NateWr: Locale keys are not automatically available in the frontend, because it would be too expensive to send all of them to the client. You have two options:

1. Use `$templateMgr->setLocaleKeys()` to pass that locale key to the frontend. See https://docs.pkp.sfu.ca/dev/ui-library/dev/#/pages/localization.

2. Pass the translated string to the component as state to the page component and replace the total in the browser:

```php
$templateMgr->setState([
    'i18nDescription' => __('admin.jobs.failed.totalCount'),
]);
```

```js
export default {
    data() {
        return {
            i18nDescription: '',
        };
    },
    computed: {
        description() {
            return this.replaceLocaleParams(
                this.i18nDescription,
                {total: this.total}
            );
        }
    }
}
```

The first method is easier. Over time I've come to prefer the second method, because it's very hard to tell exactly which component uses which locale key. But either one works.

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,97 @@
...
+		description() {
+			return this.__('admin.jobs.failed.totalCount', {total: this.total});
+		}
+	},
+	methods: {
+		handlePagination(page) {
+			this.isLoadingItems = true;
+			this.loadList(page);
+		},
+		updateJobList(data) {
```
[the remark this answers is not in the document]
touhidurabir: updated .

src/components/Container/FailedJobsPage.vue
touhidurabir: done .

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,119 @@
...
+			$.ajax({
+				url: this.apiUrl,
+				type: 'GET',
+				headers: {
+					'X-Csrf-Token': pkp.currentUser.csrfToken,
+				},
+				data: {page: page},
+				error: this.ajaxErrorCallback,
+				success: (response) => {
+					console.log(response);
```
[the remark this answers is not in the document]
touhidurabir: removed, sorry I left it .

src/components/Container/FailedJobsPage.vue
```diff
@@ -0,0 +1,119 @@
...
+import Page from './Page.vue';
+import Pagination from '@/components/Pagination/Pagination.vue';
+import PkpTable from '@/components/Table/Table.vue';
+import TableCell from '@/components/Table/TableCell.vue';
+import ButtonRow from '@/components/ButtonRow/ButtonRow.vue';
+import ajaxError from '@/mixins/ajaxError';
+
+export default {
+	name: 'FailedJobsPage',
+	extends: Page,
```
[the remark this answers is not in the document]
touhidurabir: Thats a great idea, done .

PR #269, opened 2023-03-21
add disabling option form text field

ipula:
adding disabling function for form text field
disablingOptIn


src/components/Form/fields/FieldText.vue
```diff
@@ -90,6 +88,7 @@ export default {
 	props: {
 		inputType: String,
 		optIntoEdit: Boolean,
+		disableFeild: Boolean,
```
asmecher, line 91: Should be `disableField` (here and elsewhere)

src/components/Form/fields/FieldText.vue
```diff
@@ -80,7 +78,7 @@
 		</div>
 	</div>
 </template>
-
+optIntoEdit
```
asmecher, line 81: This looks like left-over debugging text.

src/components/Form/fields/FieldText.vue
```diff
@@ -52,8 +52,6 @@
 				/>
 				<span
 					v-if="prefix"
-					class="pkpFormField__inputPrefix"
```
asmecher, line 55: Are these two lines removed for a reason?

asmecher (changes requested):
Hi @ipula! Here are a couple of initial comments -- once these comments are addressed I'll pass it along to someone else who knows ui-library better than me for a second round of review.

src/components/Form/fields/FieldText.vue
ipula: I think this also merging issue

src/components/Form/fields/FieldText.vue
ipula: I will check I think merging issue

src/components/Form/fields/FieldText.vue
ipula: I wll fix this and recommit

PR #270, opened 2023-03-22
pkp/pkp-lib#8734 Removed extra process call to handle bulk email sending

touhidurabir:
PR for pkp/pkp-lib#8734

### DIFF
--- a/src/components/Container/AccessPage.vue
+++ b/src/components/Container/AccessPage.vue
@@ -1,48 +1,19 @@
 <script type="text/javascript">
 import Page from './Page.vue';
 import NotifyUsersForm from '@/components/Form/context/NotifyUsersForm.vue';
-import ProgressBar from '@/components/ProgressBar/ProgressBar.vue';
 
 export default {
 	name: 'AccessPage',
 	extends: Page,
 	components: {
 		NotifyUsersForm,
-		ProgressBar,
 	},
 	data() {
 		return {
-			completedJobs: 0,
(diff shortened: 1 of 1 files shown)


src/components/Container/AccessPage.vue
```diff
@@ -1,48 +1,19 @@
...
 		NotifyUsersForm,
-		ProgressBar,
 	},
 	data() {
 		return {
-			completedJobs: 0,
-			progressUrl: '',
-			queueId: '',
-			totalJobs: 0,
+			totalBulkJobs: 0,
```
Vitaliy-1, line 13: Is this data used anywhere? I think a total number of jobs was here to inform about the progress of processing jobs.

touhidurabir: Thats has used there as an reference to confirm that all jobs has been pushed to queue successfully and display following view to confirm users 

CleanShot 2023-03-25 at 16 57 53@2x

we can also just pass a simple boolean flag to to confirm and present the view also and I think that better as passing number of batches make no sense there .

touhidurabir: Thinking about it again , passing a boolean flag or number to successfully dispatched batches don't make much difference just to show the success view through vue

Vitaliy-1: Yes, you are right, I see it now. The concept on how the queue is processed has changed here and the form likely should be reconsidered but for now it looks good. Thanks!

PR #275, opened 2023-07-05
Added 'Same institution as author' badge to reviewer select modal

MrRob100:
Screenshot 2023-08-15 at 10 04 17
Screenshot 2023-08-15 at 10 02 54
Screenshot 2023-08-15 at 10 01 31

Adding 'Same institution as author' badge to the list of potential reviewers during reviewer selection

The rest of the feature is in lib-pkp here https://github.com/pkp/pkp-lib/pull/9090


src/components/Container/AdvancedSearchReviewerContainer.vue
```diff
@@ -0,0 +1,62 @@
...
+					v-for="(affiliations, author) in components.selectReviewer.authors"
+					:key="author"
+				>
+					<span>{{ author }}</span>
+					<span v-if="affiliations"> - </span>
+					<span>{{ affiliations }}</span>
+				</div>
+			</div>
+			<div v-if="Object.keys(components.selectReviewer.authors).length > 4">
+				<button
```
jardakotesovec, line 17: This is really minor suggestion. In Vue.js usually more complex expressions are usually done via computed properties. And in template you would refer computed property.

src/components/Container/AdvancedSearchReviewerContainer.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+	<div>
+		<div class="section">
+			<h3>{{ components.selectReviewer.submissionAuthorList }}</h3>
+			<div :class="authorContainerclasses">
+				<div
```
jardakotesovec, line 6: From accessibility point of view its lot better to express lists semantically via ul&li tags, so its well interpreted by screen readers.

src/components/Container/AdvancedSearchReviewerContainer.vue
```diff
@@ -0,0 +1,62 @@
...
+					class="author_row"
+					v-for="(affiliations, author) in components.selectReviewer.authors"
+					:key="author"
+				>
```
jardakotesovec, line 10: I have not tested the code so I might be wrong - but this looks like it will always render all authors regardless whether show more is clicked?

jardakotesovec (changes requested):

MrRob100:
@jardakotesovec These are in now

jardakotesovec:
@Devika008 @Vitaliy-1 @asmecher 

Hi! Currently testing how it behaves in OMP. And it does show list of authors in both modals, when adding reviewer for both Internal and External reviews. Is that desirable?

Screenshot 2023-09-06 at 9 55 53


Also for some reason the detection whether reviewer is from same university as one of the authors is not working on OMP.. will check why is that.

PR #279, opened 2023-08-03
pkp/pkp-lib#8980 Add fallbacks for float inline-start due limited bro…

jardakotesovec:
…wser support.

### DIFF
--- a/src/components/Form/FormGroup.vue
+++ b/src/components/Form/FormGroup.vue
@@ -215,15 +215,25 @@ export default {
 }
 
 .pkpFormGroup__heading {
-	float: inline-start;
+	float: left;
 	width: 30%;
 	padding-inline-end: 1.5rem;
 	line-height: @line-sml;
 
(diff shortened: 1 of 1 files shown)


src/components/Form/FormGroup.vue
```diff
@@ -215,18 +215,35 @@ export default {
 }
 
 .pkpFormGroup__heading {
+	/* fallback as support float inline-start is quite limited */
```
Vitaliy-1, line 218: I'd prefer one general comment with a brief explanation and a link to an issue, like `pkp/pkp-lib#8980`, rather than mentioning it every encounter. It looks like these property values occur only in this one file.

Vitaliy-1 (commented):
Looking at the cross browser support for the `inline-start` and `inline-end`, as you already mentioned, we can completely drop their usage. It doesn't look like they will be better supported in the future. But leaving it for you to decide.

PR #280, opened 2023-08-22
Change submission language

jyhein:
- Change the submission language after submission
- Show the language of the submission in the workflow page and the author dashboard
- In the modal window editors can change the language and add/edit required metadata
- The submission's metadata can be edited even if the submission's locale is not one of the currently supported submission locales
- The language of the submission can not be changed if the status of the publication is published or there are more than one version

### DIFF
--- a/src/pages/workflow/ChangeSubmissionLanguage.vue
+++ b/src/pages/workflow/ChangeSubmissionLanguage.vue
@@ -1,0 +1,52 @@
+<template>
+	<SideModalBody>
+		<template #pre-title>
+			{{ props.submissionId }}
+		</template>
+		<template #title>
+			{{ t('submission.list.changeSubmissionLanguage.title') }}
+		</template>
+		<template #description>
+			{{ store.publicationTitle }}
+		</template>
+		<div class="p-4">
+			<div class="bg-secondary p-4">
+				<div id="changeSubmissionLanguage">
+					<PkpForm
+						v-bind="store.form"
+						@set="store.setCustom"
+						@success="store.success"
+						@cancel="store.closeSideModal"
+					></PkpForm>
+				</div>
+			</div>
+		</div>
+	</SideModalBody>
+</template>
+
+<script setup>
+import {defineProps} from 'vue';
+import PkpForm from '@/components/Form/Form.vue';
+import SideModalBody from '@/components/Modal/SideModalBody.vue';
+import {useChangeSubmissionLanguageStore} from '@/pages/workflow/changeSubmissionLanguageStore';
+
+const props = defineProps({
+	form: {
+		type: Object,
+		required: true,
+	},
+	publicationId: {
+		type: Number,
+		required: true,
+	},
+	submissionId: {
+		type: Number,
+		required: true,
+	},
+});
+const store = useChangeSubmissionLanguageStore({
+	form: props.form,
+	publicationId: props.publicationId,
+	submissionId: props.submissionId,
+});
+</script>
--- a/src/pages/workflow/changeSubmissionLanguageStore.js
+++ b/src/pages/workflow/changeSubmissionLanguageStore.js
@@ -1,0 +1,115 @@
+import {inject, ref} from 'vue';
+import {defineComponentStore} from '@/utils/defineComponentStore';
+import {useApiUrl} from '@/composables/useApiUrl';
+import {useFetch} from '@/composables/useFetch';
+import {useForm} from '@/composables/useForm';
+import {useLocalize} from '@/composables/useLocalize';
+import cloneDeep from 'clone-deep';
+
+export const useChangeSubmissionLanguageStore = defineComponentStore(
+	'changeSubmissionLanguage',
+	(props) => {
+		/**
+		 * Variables and init data
+		 */
+
+		const {t, tk} = useLocalize();
+
+		const DescriptionLocaleKeys = {
+			title: tk(
+				'submission.list.changeSubmissionLanguage.metadataDescription.title',
+			),
+			abstract: tk(
+				'submission.list.changeSubmissionLanguage.metadataDescription.abstract',
+			),
+		};
+
+		const {
+			apiUrl: {value: publicationApiUrl},
+		} = useApiUrl(
+			`submissions/${props.submissionId}/publications/${props.publicationId}`,
+		);
+
+		const {
+			form: {value: form},
+			getValue,
+			set,
+			setValue,
+		} = useForm(cloneDeep(props.form));
+		// Set action api url
+		form.action = publicationApiUrl + '/changeLocale';
+
+		// Set initial value
+		const publicationTitle = ref(getValue('title'));
+
+		const publicationProps = {};
+		// Get publication props
+		getData();
+
+		const closeModal = inject('closeModal');
+
+		/**
+		 * Functions
+		 */
+
+		function closeSideModal() {
+			closeModal();
+		}
+
+		/**
+		 * Set form data
+		 */
+		const setCustom = (_, data) => {
+			set(_, data);
+			const oldLocale = form.primaryLocale;
+			const newLocale = getValue('locale');
+			// Set fields when changing language
+			if (newLocale !== oldLocale) {
+				form.primaryLocale = newLocale;
+				form.fields.forEach((field) => {
+					if (publicationProps[field.name]) {
+						setValue(
+							field.name,
+							publicationProps[field.name][newLocale] ??
+								publicationProps[field.name],
+						);
+						field.description = t(DescriptionLocaleKeys[field.name], {
+							language: getLocaleName(newLocale),
+						});
+					}
+				});
+			}
+		};
+
+		/**
+		 * Form success
+		 */
+		const success = () => {
+			window.location.reload();
+		};
+
+		return {closeSideModal, setCustom, success, form, publicationTitle};
+
+		/**
+		 * Aux functions
+		 */
+
+		async function getData() {
+			const {data, fetch} = useFetch(publicationApiUrl, {
+				method: 'GET',
+			});
+			await fetch();
+
+			Object.assign(publicationProps, data.value ?? {});
+			delete publicationProps['locale'];
+
+			publicationTitle.value = data.value.title[props.form.primaryLocale];
+		}
+
+		function getLocaleName(locale) {
+			return form.fields
+				.find(({name}) => name === 'locale')
+				.options.find(({value}) => value === locale).label;
+		}
+	},
+);
--- a/src/components/Container/WorkflowPage.vue
+++ b/src/components/Container/WorkflowPage.vue
@@ -9,12 +9,14 @@ import PkpHeader from '@/components/Header/Header.vue';
 import LocalizeSubmission from '@/mixins/localizeSubmission.js';
 import ajaxError from '@/mixins/ajaxError';
 import dialog from '@/mixins/dialog.js';
+import ChangeSubmissionLanguage from '@/pages/workflow/ChangeSubmissionLanguage.vue';
 import SelectRevisionDecisionModal from '@/pages/workflow/SelectRevisionDecisionModal.vue';
 import {useModal} from '@/composables/useModal';
 
 export default {
 	name: 'WorkflowPage',
 	components: {
+		ChangeSubmissionLanguage,
 		ContributorsListPanel,
 		Composer,
 		Dropdown,
@@ -28,6 +30,8 @@ export default {
 		return {
 			activityLogLabel: '',
 			canAccessPublication: false,
+			canChangeSubmissionLanguage: false,
+			currentSubmissionLanguageLabel: '',
 			canEditPublication: false,
 			currentPublication: null,
 			decisionUrl: '',
@@ -309,6 +313,20 @@ export default {
 					'" ' +
 					'class="pkp_modal pkpModalWrapper" tabIndex="-1"></div>',
 			).pkpHandler('$.pkp.controllers.modal.AjaxModalHandler', opts);
+		},
+
+		/**
+		 * Open a modal displaying the change submission language form
+		 */
+		openChangeSubmissionLanguageModal() {
+			const {openSideModal} = useModal();
+			openSideModal(ChangeSubmissionLanguage, {
+				form: this.components[
+					pkp.const.FORM_CHANGE_SUBMISSION_LANGUAGE_METADATA
+				],
+				publicationId: this.workingPublication.id,
+				submissionId: this.submission.id,
+			});
 		},
 
 		/**
@@ -732,6 +750,20 @@ export default {
 			animation-duration: 0.8s;
 		}
 	}
+}
+
+.pkpSubmission__localeNotSupported {
+	margin: 0 -2rem;
+	padding: 1rem;
+	background: @primary;
+	font-size: @font-sml;
+	color: #fff;
+	text-align: center;
+}
+
+.pkpPublication__changeSubmissionLanguage {
+	display: block;
+	padding-bottom: 0.25rem;
 }
 
 // Integrate the grids in the publication tab
--- a/src/components/Form/Form.vue
+++ b/src/components/Form/Form.vue
@@ -369,14 +369,9 @@ export default {
 						missingValue = !value;
 						break;
 					case 'string':
-					case 'array':
-						if (!value.length) {
-							missingValue = true;
-						}
-						break;
 					case 'object':
 						// null values are stored as objects
-						if (!value) {
+						if (!value || (Array.isArray(value) && !value.length)) {
 							missingValue = true;
 						}
 						break;
--- a/src/components/Form/formHelpers.js
+++ b/src/components/Form/formHelpers.js
@@ -38,6 +38,9 @@ export function shouldShowGroup(group, fields) {
 	if (typeof group.showWhen === 'string') {
 		return !!whenField.value;
 	}
+	if (Array.isArray(group.showWhen[1])) {
+		return group.showWhen[1].includes(whenField.value);
+	}
 	return whenField.value === group.showWhen[1];
 }
 


bozana:
Hi @jardakotesovec, could you please take a look at this PR?
Thanks a lot!

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+						@success="changeLanguage"
+					></pkp-form>
+				</div>
+			</div>
+		</div>
+	</SideModalBody>
+</template>
+
+<script type="text/javascript">
+import {inject, reactive} from 'vue';
```
jardakotesovec, line 27: We intend to follow vue.js recommendations by default (unless we have some good reason to do otherwise). In this case we use the <script setup> syntax which is recommended when using SFC - https://vuejs.org/guide/scaling-up/sfc.html .

What that means in practice is really couple of adjustments, like using defineProps, which has the same syntax for props, but its using this macro - https://vuejs.org/guide/components/props.html


Good template to follow is `pages/reviewerSubmission/ReviewerSubmissionPage*` files, which has very similar situation of creating modal.

So please also move the business logic to the separate store component - which is really just moving it, as the store follow the component lifecycle and use exact same composition api. But it makes easier from plugin to interact with it rather than when its directly in component - more details is here - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-pinia-store--docs .
-

But its mostly about just moving things around as the core idea of using composition api is the same.

And once the whole UI would be simplified (less api interactions and workflow interactions) - it also will be lot easier to add it to storybook.

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+import cloneDeep from 'clone-deep';
+
+export default {
+	name: 'ChangeSubmissionLanguage',
+	components: {
+		PkpForm,
+		SideModalBody,
+	},
+	props: {
+		contributors: {
```
jardakotesovec, line 41: Number of props after the adjustments should hopefully drop significantly.

There is good benefit to have the modals as independent as possible from main screen and keep the communication between them at the minimum - as that makes easier for both to evolve independently.

If we manage to move the updating meta to the API endpoint, instead of handling the whole process from the UI - I think we will get just to passing the cslmform and submission. It could be even more independent and just pass submissionId & cslmform and fetch the submission from the API.

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+			type: String,
+			required: true,
+		},
+		submissionFileApiUrl: {
+			type: String,
+			required: true,
+		},
+	},
+	emits: ['set', 'updated:contributors', 'updated:files', 'updated:submission'],
+	setup(props, {emit}) {
```
jardakotesovec, line 71: I think it has such a large impact on underlying workflow page and its going to be used quite rarely - I would recommend just to reload the page once the form is succesfully submitted to reload the workflow page, rather than trying to update everything. That should simplify things significantly - and thats always good thing :-).

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+						message: r.responseJSON?.errorMessage ?? t('common.unknownError'),
+						actions: [{label: t('common.ok'), callback: (close) => close()}],
+					});
+				}
+			}
+
+			async function send(url, data, method = 'POST') {
+				if (!url) return Promise.resolve({});
+				return $.ajax({
+					url: url,
```
jardakotesovec, line 199: With new composition API I created couple of composables to make various things easier.  Please follow the useFetch examples here - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/composables-usefetch--docs

It will handle many things for you, things like passing csrf token and adding the x-http-method-override logic and error handling for unexpected errors.

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+			closeModal();
+			return Promise.resolve(1);
+
+			/**
+			 * Contributor props
+			 */
+			async function editContributors() {
+				const url = (id) => `${props.contributorsApiUrl}/${id}`;
+				await editSendEmit(
+					props.contributors,
```
jardakotesovec, line 123: Created another composable for creating api urls, so we don't have to constantly passing them from PHP as configuration - check out - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/composables-useapiurl--docs

src/components/Container/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,310 @@
...
+					}),
+				);
+				return updatedItems;
+			}
+
+			/**
+			 * Update and emit new form
+			 */
+			function updateForm(updatedSubmission) {
+				const form = cloneDeep(state.form);
```
jardakotesovec, line 242: The form thats being presented in mockups is not multilingual. I would suggest disable multilingual option there, so you don't have to fight it, because currently it still indicates the globe and you are using css override to hide the switcher. Overriding css should be the last resort..  as thats always asking for troubles down the road. If we start evolving Form component - it just stops working and it will be difficult to notice that.

jardakotesovec (changes requested):
Hi Jyrki,
you are on the good track - but hopefully we will be able to simplify things.

I added some suggestions how to proceed - feel free to follow up with questions either on github or mattermost.

jardakotesovec:
One more thing - we also now have useForm composable - but its not covered in storybook yet, because its still in early stages, but could be useful for your use case.

Purpose is basically to stop writing code which is traversing the form structures - and have that abstracted that out.

So for example if you would want to programtically set some values, you can do something like:

```
const { form, set, setValue } = useForm(props.changeLanguageForm);

setValue('abstract', abstract);
```

And you can use the `set` to process the changes coming from form:

```
<pkp-form
  v-bind="form"
  @set="set"
/>

```

As mentioned its early implementation - so if you run into any issues there please let me know and I will improve it..

src/components/Container/ChangeSubmissionLanguage.vue
jyhein: How do you suggest to disable multilingual option? The form is multilingual and needs to be sent as such. Adding hide-option prop to Form.vue for the switcher is simple but it will leave those globes visible.

src/components/Container/ChangeSubmissionLanguage.vue
jyhein: So, store component js does the language change but ChangeSubmissionLanguage.vue reloads the page?

src/components/Container/ChangeSubmissionLanguage.vue
jardakotesovec: Are you creating new dedicated API endpoint to trigger the language change?  Such endpoint would probably expect just simple fields based on the modals - target language, title, abstract, ... .  And I would not expect this to be sent in multi lingual data shape.  Does that sound reasonable or you had some other plan?

src/components/Container/ChangeSubmissionLanguage.vue
jardakotesovec: Store basically can have all the business logic, including the reload.

src/components/Container/ChangeSubmissionLanguage.vue
bozana: Hi @jyhein, yes, we have discussed this internally and we have decided to have a new API endpoint for this, so that all business logic can be removed from Vue.
@ewhanson, I suppose the API endpoint will go into the PKPSubmissionController? What should be the API path for this, e.g. submissions/$submissionId/publications/$publicationId/changeLocale, or submissions/$submissionId/changeLocale or .../setLocale... or .../locale... or ... ? :-)
Thanks!

ewhanson: Yes, I would add it to `PKPSubmissionController` and would put this under `submissions/{$submissionId}/saveLocale`. You can look at the `saveContributorsOrder` endpoint as a good example to follow: https://github.com/pkp/pkp-lib/blob/c3c4090d90c88fc5ce3a9877d7b0fa6d61c18300/api/v1/submissions/PKPSubmissionController.php#L260-L262. 

Another thing that will need to be considered is which roles will be allowed to make this change.

Happy to have a look at it once it's ready for review!

bozana: I think journal manager and editor can make the change -- they always have access to submisison functions, but for section editor (sub editors) we need to check if the user is assigned to the submission.

jardakotesovec:
@jyhein I see there is some good progress - let me know when its ready for review. Thank you.

src/components/Container/WorkflowPageOMP.vue
```diff
@@ -72,9 +72,13 @@ export default {
 
 		/**
 		 * Update the work type to be an edited volume
+		 * Update authors' isVolumeEditor to bool if null
 		 */
 		setAsEditedVolume() {
 			this.updateWorkType(this.getConstant('WORK_TYPE_EDITED_VOLUME'));
+			this.workingPublication.authors.forEach(
```
ewhanson, line 79: No problem with this, but just wondering if this was something that had been breaking the change submission locale functionality in some way?

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+
+export const useChangeSubmissionLanguageStore = defineComponentStore(
+	'changeSubmissionLanguage',
+	(props) => {
+		/**
+		 * Variables
+		 */
+
+		const {
+			apiUrl: {value: apiUrl},
```
ewhanson, line 16: I would specify which API URL this is since it's used both as the publication endpoint and as part of the `changeLocale` endpoint. Maybe something like `publicationApiUrl` as distinct from `changeLocaleApiUrl`?

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+
+		const {
+			form: {value: form},
+			setValue,
+		} = useForm(cloneDeep(props.form));
+		// Set action api url
+		form.action = apiUrl + '/changeLocale';
+
+		const publicationTitle =
+			props.form.fields[2].value[props.form.primaryLocale];
```
ewhanson, line 29: Is the publication title guaranteed to be index 2 in the `props.form.fields` array? This may be totally fine but just wanted to flag it.

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+
+		const publicationProps = {};
+		// Get publication props
+		(async () => {
```
ewhanson, line 33: This looks fine, but just wanted to double check (cc: @jardakotesovec) that an IIFE was beset practice for preloading this data here as I'm less familiar with working with stores.

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+		 * Set form data
+		 */
+		const set = (_, data) => {
+			Object.keys(data).forEach((key) => (form[key] = data[key]));
+			const oldLocale = form.primaryLocale;
+			const newLocale = data.fields?.[0].value ?? oldLocale;
+			// Set fields when changing language
+			if (newLocale !== oldLocale) {
+				form.primaryLocale = newLocale;
+				form.fields.forEach((f) => {
```
ewhanson, line 62: Another minor one, but for readability, I prefer having more explicit variable names as they make it easier to parse the code at a glance (maybe `field` instead of `f`?).

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+					}
+				});
+			}
+		};
+
+		/**
+		 * Form success
+		 */
+		const success = () => {
+			location.reload();
```
ewhanson, line 83: Probably another question for @jardakotesovec on house style. Elsewhere in the codebase, the `window` is explicitly referenced (so `window.location.reload()` instead of just `location.reload()`). Should we keep it consistent with the rest of the codebase and include `window` or is it better to leave it off?

src/components/Container/WorkflowPageOMP.vue
jardakotesovec: Hi @jyhein, Could you provide some details on this one? Also not sure why this was needed and whether its related to locale functionality.

src/pages/workflow/ChangeSubmissionLanguage.vue
```diff
@@ -0,0 +1,52 @@
...
+		<template #title>
+			{{ t('submission.list.changeSubmissionLanguage.title') }}
+		</template>
+		<template #description>
+			{{ store.publicationTitle }}
+		</template>
+		<div class="p-4">
+			<div class="bg-secondary p-4">
+				<div id="changeSubmissionLanguage" aria-live="polite">
+					<PkpForm
```
jardakotesovec, line 15: Interesting, have not seen use of aria-live user on form like this. Can you help me understand how it improves accessibility?

src/pages/workflow/changeSubmissionLanguageStore.js
jardakotesovec: Rather than destructing 'value'.  More common approach with composition api, is to keep it as ref and when you need to access value, you do it like: `form.value.action = apiUrl.value + '/changeLocale'` and as Erik mentioned destructing it like {apiUrl: publicationApiUrl} would make it bit more friendly.

src/pages/workflow/changeSubmissionLanguageStore.js
jardakotesovec: @ewhanson @jyhein Its bit unusual, not technically wrong though :-). Did you have any reason to wrap it in function like this? Maybe just to avoid variable name conflicts? That could be avoided for example with `const {data: publication, fetch: fetchPublication} = useFetch(..).`

Or if you would want to keep it in function, I would suggest just to create function with name and than invoke it.

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+
+		function closeSideModal() {
+			closeModal();
+		}
+
+		/**
+		 * Set form data
+		 */
+		const set = (_, data) => {
+			Object.keys(data).forEach((key) => (form[key] = data[key]));
```
jardakotesovec, line 56: My understanding here is that you want to extend `set` function, so if the language gets changed it also updates title and abstract, right?

To just to make it bit more explicit that you are extending default function. You can get the `set` function from userForm - const {set} = useForm(...).   And than create your own function `setCustom`, which calls the original set function and do your custom bit.

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+			const oldLocale = form.primaryLocale;
+			const newLocale = data.fields?.[0].value ?? oldLocale;
+			// Set fields when changing language
```
jardakotesovec, line 59: Here I think you could use `getValue('locale')` from `const {getValue} = useForm(...).` Because at this moment form structure will be already updated from the `set` call.

src/pages/workflow/changeSubmissionLanguageStore.js
jardakotesovec: This is very minor thing, but yes we can keep it consistent with rest of the code base and add window

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+
+		return {closeSideModal, set, success, form, publicationTitle};
+
+		/**
+		 * Aux functions
+		 */
+
+		function getLocaleName(locale) {
+			return form.fields[0].options.find(({value}) => value === locale).label;
+		}
```
jardakotesovec, line 94: This would be nice to also cover with some helper function from useForm, but we don't have it yet..  Its fine to leave it as it is.

src/pages/workflow/changeSubmissionLanguageStore.js
```diff
@@ -0,0 +1,96 @@
...
+					if (publicationProps[f.name]) {
+						setValue(
+							f.name,
+							publicationProps[f.name][newLocale] ?? publicationProps[f.name],
+						);
+						// i18n.js t doesn't seem to support variable as tranlation key
+						// (i18nExtractKeys.vite.js: uiLocaleKeysBackend.json), using str.replace for now.
+						f.description = f.description.replace(
+							getLocaleName(oldLocale),
+							getLocaleName(newLocale),
```
jardakotesovec, line 72: Yes, locale keys are being determined in build time, therefore it can't really understand variables.

But in this case you are always updating title and abstract, so instead of loop, you can just always apply it for title and abstract.

Or if you want to stick with the loop, you can define the locale keys for example like this:
```
const DescriptionLocaleKeys = {
title: tk`submission.list.changeSubmissionLanguage.metadataDescription.title`,
abstract: tk`submission.list.changeSubmissionLanguage.metadataDescription.abstract`
}

```

Which than can be used as `t(DescriptionLocaleKeys['title'], {language: ...})`

Let me know if there is any issue with this.

jardakotesovec:
@jyhein Hi Jyrki, I am done with code review, apologies for delay. Its looking good. Its all really just minor suggestions.

src/pages/workflow/changeSubmissionLanguageStore.js
jyhein: My idea  here was to keep the code as general as possible, in case there will be additional fields

src/components/Container/WorkflowPageOMP.vue
jyhein: I got an error that the value is expected to be boolean but that was when vue side updated contributor props. Not sure if if it is needed anymore (or at all)

src/pages/workflow/changeSubmissionLanguageStore.js
jyhein: Vue complained something about making <script setup> async (.vue file), so I had to wrap it in a function

src/components/Container/WorkflowPageOMP.vue
jardakotesovec: @jyhein Oki, can you please double check whether is needed?

src/pages/workflow/changeSubmissionLanguageStore.js
jyhein: I wanted to make apiUrl (publicationApiUrl) immutable. No other reason

src/components/Container/WorkflowPageOMP.vue
jyhein: I removed the code because I didn't see any errors OMP side when changing language. I didn't test any further

src/pages/workflow/changeSubmissionLanguageStore.js
jardakotesovec: Thats good intention for sure, I have only mixed feeling always when I see some string replacement as that feels fragile. I know its very likely to be fine, but theoretically there can be some scenario where the locale name could be substring of some word in the description in some specific language. Its unlikely, but that still would make me to prefer to use the locale keys as suggested above.

jyhein: Ok

PR #286, opened 2023-09-06
I8919 vue3&vite migration

jardakotesovec:


src/components/Button/Button.vue
```diff
@@ -51,7 +51,6 @@ export default {
 			if (this.element === 'button') {
 				e.preventDefault();
 			}
-			this.$emit('click');
 		},
```
jardakotesovec, line 54: this was causing double click

src/components/Button/Button.vue
```diff
@@ -4,7 +4,7 @@
 		class="pkpButton"
 		:class="classes"
 		:href="element === 'a' ? href : false"
-		:disabled="isDisabled"
+		:disabled="element === 'a' ? undefined : isDisabled"
 		@click="click"
```
jardakotesovec, line 8: This is because vue3 knows that disabled is not valid for <a> element and as result it does not treat it in special way. Whats specific about disabled - that browser consider element being disabled just by being present in dom, regardless what value it has. So using undefined prevents disabled to appear in DOM.

src/components/Container/StatsEditorialPage.vue
```diff
@@ -182,11 +182,11 @@ export default {
 						row.total = total.value;
 					}
 					if (
-						averagesResponse.hasOwnProperty(row.key) &&
+						Object.prototype.hasOwnProperty.call(averagesResponse, row.key) &&
 						averagesResponse[row.key] > -1 &&
```
jardakotesovec, line 186: This is coming from newer version of eslint rules -  new rule which recommends using Object.prototype, because objects don't guarantee to have hasOwnProperty anymore if they are created with Object.create(null).

src/components/Table/Table.vue
```diff
@@ -39,10 +39,10 @@
...
 		<tbody
-			@keydown.35.ctrl.exact.prevent="focusEnd"
-			@keydown.36.ctrl.exact.prevent="focusStart"
-			@keydown.35.exact.prevent="focusLastCell"
-			@keydown.36.exact.prevent="focusFirstCell"
+			@keydown.end.ctrl.exact.prevent="focusEnd"
+			@keydown.home.ctrl.exact.prevent="focusStart"
+			@keydown.end.exact.prevent="focusLastCell"
+			@keydown.home.exact.prevent="focusFirstCell"
 			@keydown.left.exact.prevent="focusPreviousCell"
```
jardakotesovec, line 46: This came from eslint, looks good.

src/components/Form/fields/FieldUpload.vue
```diff
@@ -202,6 +209,28 @@ export default {
 		 * @return {Object}
 		 */
 		dropzoneOptions() {
+			console.log({
```
jonasraoni, line 212: Maybe forgotten debug code.

src/components/Form/fields/FieldUpload.vue
```diff
@@ -345,21 +374,29 @@ export default {
...
-		this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
-			'aria-describedby',
-			this.describedByIds
-		);
+		this.isComponentMounted = true;
+
+		// not ideal, but first this component needs to get mounted
+		// than vue-dropzone gets mounted and afterwards vue-dropzone
+		// DOM can be manipulated
+		setTimeout(() => {
```
jonasraoni, line 382: I'm guessing the `nextTick` didn't work here 😁

vite.config.js
```diff
@@ -0,0 +1,17 @@
+import path from 'path';
+import vue from '@vitejs/plugin-vue';
+
+export default {
+	base: process.env.PKP_DOCS_VERSION
+		? '/dev/ui-library/' + process.env.PKP_DOCS_VERSION
+		: '/',
+	plugins: [vue()],
+	/*optimizeDeps: {
```
jonasraoni, line 9: Perhaps it can be removed.

src/mixins/global.js
```diff
@@ -26,7 +26,7 @@ export default {
 		 * @param {Object} params (Optional) Variables to compile with the translation
 		 * @return {String}
 		 */
-		__: function (key, params) {
+		t: function (key, params) {
```
jonasraoni, line 29: I'd rather keep the `__`, it would also decrease the noise, but if there's a reason, nothing against 😁

src/docs/pages/localizeMoment.md
```diff
@@ -7,13 +7,11 @@ import moment from 'moment';
...
-        const timeSince = moment(timestamp)
-            .locale(
-                this.getMomentLocale($.pkp.app.currentLocale)
-            )
-            .fromNow();
-    }
-}
+	mixins: [localizeMoment],
+	created() {
+		const timeSince = moment(timestamp)
```
jonasraoni, line 12: Moment received a lot of hate these last years, I'm surprised it's still here 😁

src/components/Container/StatsEditorialPage.vue
jonasraoni: I see it created a lot of noise. As we know what's coming in, I'd rather just disable this rule at this moment, but now that it's done, I think it's ok to keep (considering that getting back is unsafer).

jonasraoni (approved):

src/components/Container/StatsEditorialPage.vue
jardakotesovec: @jonasraoni Yeah.. it was kind of on the edge - being small enough to include it and being annoying. I think it was around 20 cases. In retrospect I would disable it given how many details is already included in this PR.

src/components/Form/fields/FieldUpload.vue
jardakotesovec: Good catch, thank you.

vite.config.js
jardakotesovec: Yes, thank you.

src/components/Form/fields/FieldUpload.vue
jardakotesovec: Yeah, this is my least part of the upgrade.. hate these setTimeout workarounds..   Likely getting rid of this when looking for alternative to vue-dropzone in https://github.com/pkp/pkp-lib/issues/9367 .

nextTick is Node thing, is not it? Not really available in browser..

PR #290, opened 2023-10-12
pkp/pkp-lib#9408 Permit escaping of mixed content when localizing strings

asmecher:


src/mixins/global.js
```diff
@@ -146,11 +146,18 @@ export default {
 		 *
 		 * @param {String} str String to replace params in
 		 * @param {Object} params Key/value hash of params to replace
+		 * @param {bool} applyEscaping True iff replaced content should be HTML escaped
 		 * @return {String}
 		 */
-		replaceLocaleParams(str, params) {
+		replaceLocaleParams(str, params, applyEscaping = false) {
 			for (var param in params) {
```
jardakotesovec, line 153: Better to have options object as 3rd argument `{ htmlEscapling: true/false }`, as replaceLocaleParams('', {}, true) is very cryptic.

src/mixins/global.js
```diff
@@ -146,11 +146,18 @@ export default {
...
-				let value = params[param];
+				var value = params[param];
+				if (applyEscaping) {
+					var text = document.createTextNode(value);
+					var p = document.createElement('p');
+					p.appendChild(text);
+					value = p.innerHTML;
+				}
 				// If a locale object is passed, take the value from the current locale
```
jardakotesovec, line 161: I think this approach is fine.. technically can be reduced to

```
const p = document.createElement('p');
p.innerText = str;
value = p.innerHTML;
```

asmecher: Done, thanks!

src/mixins/global.js
asmecher: Done!

PR #309, opened 2024-01-08
pkp/pkp-lib#8248 COUNTER R5 TSV reports

bozana:
s. https://github.com/pkp/pkp-lib/issues/8248

### DIFF
--- a/src/components/Form/counter/CounterReportForm.vue
+++ b/src/components/Form/counter/CounterReportForm.vue
@@ -1,0 +1,137 @@
+<script>
+import PkpForm from '@/components/Form/Form.vue';
+
+export default {
+	name: 'CounterReportForm',
+	extends: PkpForm,
+	components: {
+		PkpForm,
+	},
+	props: {
+		reportId: {
+			type: String,
+			required: false,
+		},
+		reportFields: {
+			type: Array,
+			required: true,
+		},
+		earliestDate: {
+			type: String,
+			required: true,
+		},
+		lastDate: {
+			type: String,
+			required: true,
+		},
+	},
+	methods: {
+		/**
+		 * Get the report parameters
+		 *
+		 * @param Object
+		 * @return Object
+		 */
+		getReportParams(formSubmitValues) {
+			let params = {};
+			for (const [key, value] of Object.entries(formSubmitValues)) {
+				switch (key) {
+					case 'customer_id':
+					case 'begin_date':
+					case 'end_date':
+					case 'yop':
+					case 'item_id':
+						if (value != null && value.length > 0) {
+							params[key] = value;
+						}
+						break;
+					case 'metric_type':
+					case 'attributes_to_show':
+						if (value != null && value.length > 0) {
+							params[key] = value.join('|');
+						}
+						break;
+					case 'include_parent_details':
+						if (value == true) {
+							params.include_parent_details = 'True';
+						}
+						break;
+					case 'granularity':
+						if (value == true) {
+							params.granularity = 'Totals';
+						}
+						break;
+				}
+			}
+			return params;
+		},
+
+		/**
+		 * Submit the form
+		 */
+		submit() {
+			if (!this.canSubmit) {
+				return false;
+			}
+
+			let errors = this.validate();
+			if (Object.keys(errors).length) {
+				this.$emit('set', this.id, {
+					errors: {
+						...this.errors,
+						...errors,
+					},
+				});
+				return;
+			}
+
+			this.isSaving = true;
+
+			$.ajax({
+				context: this,
+				method: this.method,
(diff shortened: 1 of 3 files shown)


src/components/Form/counter/CounterReportForm.vue
```diff
@@ -0,0 +1,141 @@
...
+					errors: {
+						...this.errors,
+						...errors,
+					},
+				});
+				return;
+			}
+
+			if (this.action === 'emit') {
+				this.$emit('success', this.submitValues);
```
jardakotesovec, line 89: Do we have scenario where we actually want this particular form "just emit"?

src/components/Form/counter/CounterReportForm.vue
```diff
@@ -0,0 +1,141 @@
...
+			} else {
+				this.isDownloadingReport = true;
+
```
jardakotesovec, line 92: I think setting this.isSaving instead would show the progress indicator, as this is what FormPage.vue is using to decide whether to show spinning wheel.

src/components/Form/counter/CounterReportForm.vue
```diff
@@ -0,0 +1,141 @@
...
+					error(r) {
+						this.isDownloadingReport = false;
+						if (r.status && r.status === 400) {
+							if (r.responseJSON.hasOwnProperty('Code')) {
+								// COUNTER speific errors should actually not occur
+								// because of the form/user input validation
+								// but consider them for any case as well.
+								pkp.eventBus.$emit(
+									'notify',
+									r.responseJSON.Code +
```
jardakotesovec, line 110: This is interesting. I thought that all network related errors would go to that ajaxErrorCallback, which shows dialog. But in this case it would use notification mechanism to show the error. 

Do you have any insight to why there would be two mechanisms to display unexpected error?

jardakotesovec (commented):
Looks good, I had some troubles while testing that - I shared that on mattermost.

But from code point of view just small suggestions/questions. 

THank you

src/components/Form/counter/CounterReportForm.vue
bozana: No, I just saw it in Form.vue, so I was not sure if we need it -- I think I even do not know when exactly this is happening (for Form.vue)... :thinking:

src/components/Form/counter/CounterReportForm.vue
bozana: But it would tell "Saving", no? It is maybe also OK... What would you suggest, to use isSaving or nothing?

src/components/Form/counter/CounterReportForm.vue
bozana: I have seen this is used in Form.vue, so I used it too :-)
I am not sure if it has something to do with the forms, or maybe how the (form) errors were displayed earlier... Hmmm... :thinking: 
Maybe @ewhanson would know?

ewhanson: I think in the past I've used both, depending on whether we needed to grab the user's attention with a dialog or if we just needed to present some more passive information about what happened, but I can't remember the specifics anymore. I think maybe in the case of DOIs, deposit errors in 3.3 and earlier used the notification mechanism so I may have simply carried that forward, but I'm not sure.

src/components/Form/counter/CounterReportForm.vue
bozana: OK, I am using isSaving... :-)

src/components/Form/counter/CounterReportForm.vue
jardakotesovec: My guess is what this is used if you don't want to form handle submitting itself - but you would want to do it yourself. So instead of extending - you could just set action to 'emit' and get the data from @success event and than try to submit it. And thats actually something that would make sense in your use case, because now you are basically just customising submit method.
But I suspect that there would be small glitch, as it sets isSaving to `true` when emitting. And if you would get errors, I don't see easy way to reset the isSaving to false.

So I would suggest here just delete it from your version, as its not needed and stick with this pattern of extending original Form. As for 3.4 we don't want to do much changes to avoid regressing some use case that we are not aware of.

And for 3.5 we can get creative if we have time. With composition api we are moving much more towards composition over inheritance. And the path for future forms in composition api is likely shifting more towards having the logic inside composable (useForm and similar) rather then directly in the Form component. That eventually should give us good flexibility how to handle special cases.

src/components/Form/counter/CounterReportForm.vue
jardakotesovec: Oki.. lets stick with whats in Form.vue for now, and I will revise that once I get more to Forms in future.

PR #320, opened 2024-01-31
I9421 v-pkp-allowed-html

jardakotesovec:


src/directive/allowedHtml.js
```diff
@@ -0,0 +1,40 @@
+import DOMPurify from 'dompurify';
+
+const allowed_html =
+	'a[href|target|title],em,strong,cite,code,ul,ol,li[class],dl,dt,dd,b,i,u,img[src|alt],sup,sub,br,p';
```
jonasraoni, line 4: I guess this will be fed from the back end later, but just flagging to not be forgotten 😁

src/directive/allowedHtml.js
```diff
@@ -0,0 +1,40 @@
...
+	ALLOWED_TAGS.push(tag);
+
+	if (parts.length > 1) {
+		const attributes = parts[1].substring(0, parts[1].length - 1).split('|');
+		attributes.forEach((attribute) => ALLOWED_ATTR.push(attribute));
+	}
+});
+
+// DOMPurify does not support defining tags along with attributes, its separate lists
+// https://github.com/cure53/DOMPurify/issues/272
```
jonasraoni, line 21: There's a sample implementation here (I just skimmed the code, and it seems to be ok):
https://www.linkedin.com/feed/update/urn:li:activity:7034210540885282816/

jonasraoni (approved):

src/directive/allowedHtml.js
jardakotesovec: @jonasraoni Thanks, I have seen it. But I don't think it makes that much of the difference as its meant to be for security rather than for validation. And customising it also requires more in depth testing (ideally creating some test coverage) and also check performance characteristics.

src/directive/allowedHtml.js
jardakotesovec: @jonasraoni I was wondering whether it is worth it...  but its probably quite easy.. will add it..

jardakotesovec: Actually.. now I realised that it has to be passed as its something people might want customise..

PR #324, opened 2024-02-06
pkp/pkp-lib#9421 replace v-html with v-strip-unsave-html

jardakotesovec:


src/directives/stripUnsafeHtml.js
```diff
@@ -0,0 +1,25 @@
+import DOMPurify from 'dompurify';
+
+const sanitizeConfig = {
+	USE_PROFILES: {html: true},
+};
```
jonasraoni, line 5: I remember about leaving a comment here to use the configuration from the application, was it dropped?!

jonasraoni (approved):
@jardakotesovec Just left a question 😁

src/directives/stripUnsafeHtml.js
jardakotesovec: For 3.3 and 3.4 yes keeping things just plain and simple and safe :-).

 For 3.5 we still will iterate with Alec on some options.  Thing is most v-html are for translations (seemed like 80%) and we don't have very well defined tags and attributes for these.  So we might want do that first. 

So basically just need bit more time to figure out whether some more fine-tuned approach would be beneficial and based on that there will be separate PRs for 3.5.

PR #326, opened 2024-02-08
pkp/pkp-lib#9453 rename useTranslation to useLocalize and add getMome…

jardakotesovec:
…ntLocale and localizeSubmission to useLocalize


jardakotesovec:
Just waiting for tests to pass.

PR #328, opened 2024-02-13
pkp/pkp-lib#9527 Mark for review

jardakotesovec:


jonasraoni (approved):
FYI I've checked this yesterday and I think everything makes sense, it's not covering all changes (regarding Vue 3), and I think it shouldn't, as the official documentation is really the best guide. I just see grammar typos, but it's better for a native to get them fixed :)

jardakotesovec:
@jonasraoni Great, thanks for checking!

PR #331, opened 2024-02-16
pkp/pkp-lib#9736 allow only sorting in FieldOptions 

bozana:
s. https://github.com/pkp/pkp-lib/issues/9736


bozana:
Hi @jardakotesovec, I took your PR/changes and added some more information for the StoryBook. Could you please take a look? Thanks a lot!

jardakotesovec:
@bozana Looking good, thank you for adding that to storybook.

I might rework the markup bit at some point for better accessibility. But no need to keep this on hold for it.  So if it works as you need - feel free to merge it.

PR #333, opened 2024-02-28
pkp/pkp-lib#5885 updated and added new UI component

touhidurabir:
for pkp/pkp-lib#5885


touhidurabir:
Hi @jardakotesovec , can you take a look at the PR now . I have removed  any extra component and have the disable/enable now as part of slider component . But just could not make few tailwind css class work so design still a bit mess .

asmecher (approved):

jardakotesovec:
@touhidurabir Hi Touhidur - I see there are some changes - can you remind me is this still something to be reviewed? If so - can you just describe briefly what scenario is this addressing? Thank you

touhidurabir:
@jardakotesovec sorry for late reply . It's now ready for merge as was in testing pahse by hosting team . About the change, when there is no entry of `numDays.....` settings name in the `journal/press/server_settings` table , it shows the label as `null days before due date` . Basically in those cases we just need to show `No reminder set` . The requested change address that issue .

PR #335, opened 2024-03-06
pkp/pkp-lib#9773 html text enhancement for rich field text

touhidurabir:
for pkp/pkp-lib#9773

### DIFF
--- a/src/components/Form/fields/FieldRichText.vue
+++ b/src/components/Form/fields/FieldRichText.vue
@@ -9,6 +9,7 @@ export default {
 			type: String,
 			required: true,
 		},
+
 		size: {
 			type: String,
 			default() {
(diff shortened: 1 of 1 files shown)


src/components/Form/fields/FieldRichText.vue
```diff
@@ -56,13 +75,28 @@ export default {
...
-					invalid_elements: 'em strong',
+					invalid_elements: this.invalidElements,
+					valid_elements: this.validElements,
 
 					// Allow pasting while stripping all styles, tags, new lines and getting only text content
 					// More details at https://www.tiny.cloud/docs/tinymce/6/copy-and-paste/
+					// Note that not all options available to 5.0+ that being used right now
 					paste_preprocess: (editor, args) => {
-						args.content = $('<div>' + args.content + '</div>').text();
+						// we need to have better control as tinymec is converting specials chars to html entities
```
asmecher, line 87: *tinymce

src/components/Form/fields/FieldRichText.vue
```diff
@@ -56,13 +75,28 @@ export default {
...
 					},
+
 					extended_valid_elements: 'b,i',
-					invalid_elements: 'em strong',
+					invalid_elements: this.invalidElements,
+					valid_elements: this.validElements,
 
 					// Allow pasting while stripping all styles, tags, new lines and getting only text content
 					// More details at https://www.tiny.cloud/docs/tinymce/6/copy-and-paste/
+					// Note that not all options available to 5.0+ that being used right now
```
asmecher, line 85: This function doesn't appear to correctly handle entities. Try for example:
- Open the `Title & Abstract` metadata form
- Edit the `Abstract` field and manually type `&amp;` into it somewhere
- Copy that section of text and paste it into the `Title` field

You'll see that the `&amp;` is unexpectedly decoded into a `&` in the title.

Likewise, try typing a raw `&` into the abstract field and copy/pasting that into the title; you'll see that the operation fails entirely.

I'm not actually sure we need to use `paste_preprocess` at all; I'm not able to replicate the behaviour described here with the XSS attack (``) described below without it.

asmecher (changes requested):
Thanks, @touhidurabir, I've added a couple notes.

src/components/Form/fields/FieldRichText.vue
touhidurabir: @asmecher I think that is the expected behaviour I am trying to set there . you see, tinymce convert any paste data into entity converting format for example , `` turned into `&lt;img src=&quot;onerror=alert(1)&quot;&gt;` it's then not possible to properly only extract the context from within encapsulated in tags . Also `&amp;` on paste turning into `&` still stored in DB as `&amp;` otherwise it will be double decoded as `&amp;amp;` . also copy pasting `&` does remain as `&` . see the attached screenshot as below 
CleanShot 2024-03-09 at 14 00 28@2x

it stored the data in DB as `BOLD2 BOLD.  &amp; some text &amp;` when copy pasting form abstract field to title field and saving without modification . 

The point is getting only the context, removing any tags/formatting/encoding  as tinymce itself enforce encoding . I think this bring better result. sorry if I am missing something here .

touhidurabir: @asmecher the main purpose of the `paste_preprocess` was to extract only the content without any tags or styles and let user apply their own style as we only allow just 5 html tags such as `b,i,u,sup,sub` . The rest of quirks are gone as I have removed the server side extra sanitization and set the `valid_elements` prop on tinymce editor . So we can remove if we don't want apply the content extraction only anymore . 

As we have discussed last time, I am removing it .

PR #336, opened 2024-03-19
pkp/pkp-lib#9790 Fix color-picker value binding

jardakotesovec:


src/components/Form/fields/FieldColor.vue
```diff
@@ -70,6 +70,7 @@ export default {
 		 * @param {String} newVal
 		 */
 		setValue: function (newVal) {
+			console.log('set value triggered');
```
jonasraoni, line 73: I think this needs to be removed later.

jardakotesovec: @jonasraoni Ops, thanks for catching that!

PR #339, opened 2024-03-21
pkp/pkp-lib#9771 Move ORCID functionality into core application

ewhanson:


ewhanson:
@jardakotesovec, would you be able to review this? This is part of the larger ORCID integration work in pkp/pkp-lib#9771. Thanks!

ewhanson:
Hey @jardakotesovec, 

Thanks for having a look. And yes, I can reproduce the issue and it was on my list of things to ask you about once this work was done! (It was also affecting the DOI settings as well).

In terms of surprises, maybe at least one more 😂

When accessing the contributor form with the ORCID functionality enabled, closing the modal for the `FieldOrcid` will also dismiss its parent modal (in this case the contributor form). You may need ORCID sandbox credentials in order to fully test this (They can be generated from sandbox.orcid.org or I can share my set with you privately). Getting up and running with the ORCID sandbox is a little involved as it only accepts mailinator emails for anything.

Here's a short screencast of the modal issue happening:


https://github.com/pkp/ui-library/assets/20844845/cf3db8f2-e39f-4f30-9eb3-a65b70412382

jardakotesovec:
@ewhanson Yes, thats known problem, will tackle it in second half of June. Created issue you can track for progress - https://github.com/pkp/pkp-lib/issues/9992

src/components/Form/fields/FieldOrcid.vue
```diff
@@ -0,0 +1,230 @@
...
+			const {data, validationError, fetch} = useFetch(apiUrl, {
+				method: 'POST',
+				expectValidationError: true,
+			});
+
+			await fetch();
+
+			if (validationError.value !== null) {
+				pkp.eventBus.$emit('notify', validationError.value['error'], 'warning');
+			}
```
jardakotesovec, line 169: What kind of validationError do you actually expect?

src/components/Form/fields/FieldOrcid.stories.js
```diff
@@ -0,0 +1,48 @@
+import FieldOrcid from '@/components/Form/fields/FieldOrcid.vue';
+import FieldBaseMock from '../mocks/field-base';
+import FieldOrcidMock from '../mocks/field-orcid';
+import {http, HttpResponse} from 'msw';
+
+export default {
```
jardakotesovec, line 6: Would be great to add couple of more stories to cover multiple states as possible. Whats great on storybook that it makes screenshot per every story - therefore more stories you do - more you have captured component state in screenshots and it helps catch any regressions.

Also having difficulties to see localisations - so that might need to be added to global.js. Hopefully get soon enough to automate this process.

src/components/Form/Form.vue
```diff
@@ -210,7 +210,7 @@ export default {
 		submitValues() {
 			let values = {};
 			this.fields.forEach((field) => {
-				if (field.component === 'field-html') {
+				if (field.isInert) {
```
jardakotesovec, line 213: Thats a nice improvement, thanks! I assume that you added enabled isInert for fieldHtml as well :-).

src/components/Form/context/OrcidSettingsForm.vue
```diff
@@ -0,0 +1,28 @@
...
+		 *
+		 * @param {String} name Name of the field to modify
+		 * @param {String} prop Name of the prop to modify
+		 * @param {mixed} value The new value for the prop
+		 * @param {String} localeKey Optional locale key for multilingual props
+		 */
+		fieldChanged: function (name, prop, value, localeKey) {
+			// TODO: This isn't actually working as the fields are properly required.
+			if (name === 'orcidEnabled') {
+				this.removeError('orcidApiType', localeKey);
```
jardakotesovec, line 20: Hopefully resolved with https://github.com/pkp/pkp-lib/issues/9996

src/components/Form/fields/FieldOrcid.stories.js
ewhanson: Thanks, that's a good idea. I'll look into what states should be reflected for this. Are there any components you think would be a good example to look at that we already have?

And I thought I'd already added the locale strings to `global.js` but I hadn't. I've done that now and updated the PR.

jardakotesovec: @ewhanson ReviewerSubmissionPage.stories.js is quite good example as it shows how you can add `play` function to the story, which for example can click on some button there, which displays the modal and the end result will be snapshotted.

Only tricky thing with modals is that chromatic does not detect their size - so the height needs to be set explicitely. Good strategy is via decorator, also can be seen in that ReviewerSubmissionPage.stories.js

src/components/Form/fields/FieldOrcid.vue
```diff
@@ -0,0 +1,230 @@
...
+			const {data, validationError, fetch} = useFetch(apiUrl, {
+				method: 'POST',
+				expectValidationError: true,
+			});
+			await fetch();
+
+			if (validationError.value !== null) {
+				this.verificationRequested = false;
+				pkp.eventBus.$emit('notify', validationError.value['error'], 'warning');
+			}
```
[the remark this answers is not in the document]
ewhanson: They're not so much validation errors, but if the API returns an error. In this (and the previous) case, it shouldn't ever really happen, but the API can return a 400-level response. The API code is here and can return a few different 400-level responses: https://github.com/ewhanson/pkp-lib/blob/1a55dbf6f54f3a06d36b4b42b4e7764094e3b39c/api/v1/orcid/OrcidController.php#L72-L123.

I'm not sure what the best approach here would be. On the one hand, they aren't really validation errors and they shouldn't ever really happen in practice, but on the other hand, I don't really want to silently ignore if the API doesn't return 200. Any thoughts on the best approach here? I suppose another option would be a generic "something went wrong" response.

src/components/Form/fields/FieldOrcid.vue
ewhanson: Same as mentioned below.

src/components/Form/Form.vue
ewhanson: Yup!

src/components/Form/fields/FieldOrcid.vue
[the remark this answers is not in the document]
jardakotesovec: Intention is covered here - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/composables-usefetch--docs

> Error Handling is implemented in useFetch. By default, any response that is not a 200 status is considered an unexpected error, and a dialog is displayed to inform the user. If you anticipate a validation error of user input (4xx status codes) that require handling, this must be explicitly enabled. For more details, refer to the [POST 4xx](https://main--6555d3db80418bb1681b8b17.chromatic.com/iframe.html?path=/docs/composables-usefetch--docs#post-4xx-with-validation-error) in the Examples section.

But maybe the wording is not clear there? If you don't expect any validation errors related to the user input that you would want to communicate back to the user - just dont add `expectValidationError` and it shows general error dialog if something happens out of the box.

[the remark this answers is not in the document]
ewhanson: Thanks, that's helpful. I think the wording is clear enough, I just don't think I read it closely enough.

src/styles/_global.less
```diff
@@ -31,3 +31,5 @@
 @import 'variables';
 @import 'helpers';
 @import 'elements/screen-reader';
+
+@import '../../../pkp/styles/orcid.less';
```
ewhanson, line 35: @jardakotesovec, this relative path to `pkp-lib` is breaking the Chromatic tests. Do you have any advice/idea on how best to handle this? The styles are used by both the themes, backend `.tpl`s and the Vue components.

ewhanson: No longer needed in the ui-library so shouldn't be a problem anymore.

PR #348, opened 2024-05-09
pkp/pkp-lib#9950 Add laravel pagination support for useFetchPaginated

jardakotesovec:


src/utils/i18n.js
```diff
@@ -16,7 +16,7 @@
 export function t(key, params) {
 	if (
 		typeof pkp === 'undefined' ||
-		typeof pkp.localeKeys[key] === 'undefined'
+		typeof pkp?.localeKeys?.[key] === 'undefined'
```
ewhanson, line 19: Was this and the other optional chaining additions causing issues previously, or was there something specific to these changes that drew your attention to this?

src/composables/useFetch.test.js
```diff
@@ -15,6 +15,7 @@ import {setActivePinia, createPinia} from 'pinia';
 
 import {useFetch} from './useFetch';
 import {useModalStore} from '@/stores/modalStore';
+global.pkp = global.pkp || {};
```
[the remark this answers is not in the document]
jardakotesovec: @ewhanson yeah, we have not had ui-library unit tests running on CI - so it regressed over time..

We are adding the it here https://github.com/pkp/pkp-lib/issues/9951

src/utils/i18n.js
jardakotesovec: It was causing issues previously.. but noticed it while working on this, when I run tests after some time.

src/composables/useFetch.test.js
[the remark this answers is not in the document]
ewhanson: Awesome, that's great!

src/utils/i18n.js
ewhanson: Figured it was something like this, but wanted to make sure I wasn't missing anything else. Thanks!

PR #350, opened 2024-05-20
Log reviewer response

MrRob100:
issue https://github.com/pkp/pkp-lib/issues/9893

Screenshot 2024-05-29 at 17 18 38

### DIFF
--- a/src/managers/ReviewerManager/useReviewerManagerActions.js
+++ b/src/managers/ReviewerManager/useReviewerManagerActions.js
@@ -4,6 +4,9 @@ import {useModal} from '@/composables/useModal';
 import {useLocalize} from '@/composables/useLocalize';
 import {useSubmission} from '@/composables/useSubmission';
 import {useFetch, getCSRFToken} from '@/composables/useFetch';
+import WorkflowLogResponseModal from '@/managers/ReviewerManager/modals/WorkflowLogResponseModal.vue';
+import {useDashboardPageStore} from '@/pages/dashboard/dashboardPageStore';
+import {useApiUrl} from '@/composables/useApiUrl';
 
 export const Actions = {
 	REVIEWER_ADD_REVIEWER: 'reviewerAddReviewer',
@@ -22,6 +25,7 @@ export const Actions = {
 	REVIEWER_THANK_REVIEWER: 'reviewerThankReviewer',
 	REVIEWER_REVERT_CONSIDER: 'reviewerRevertConsider',
 	REVIEWER_SEND_REMINDER: 'reviewerSendReminder',
+	REVIEWER_LOG_RESPONSE: 'reviewerLogResponse',
 };
 
 export function useReviewerManagerActions() {
@@ -185,6 +189,14 @@ export function useReviewerManagerActions() {
 			name: Actions.REVIEWER_EDITORIAL_NOTES,
 			icon: 'DefaultDocument',
 		});
+
+		if (!reviewAssignment.dateConfirmed) {
+			actions.push({
+				label: t('editor.review.logResponse'),
+				name: Actions.REVIEWER_LOG_RESPONSE,
+				icon: 'ReviewAssignments',
+			});
+		}
 
 		return actions;
 	}
@@ -566,6 +578,37 @@ export function useReviewerManagerActions() {
 		openLegacyModal({title: t('editor.review.reminder')}, finishedCallback);
 	}
 
+	function reviewerLogResponse(
+		{submission, reviewAssignment, submissionStageId},
+		finishedCallback,
+	) {
+		const {openSideModal} = useModal();
+
+		const dashboardStore = useDashboardPageStore();
+		let form = dashboardStore.componentForms.logResponseForm;
+		let submissionId = submission.id;
+
+		const {getCurrentPublication} = useSubmission();
+		const currentPublication = getCurrentPublication(submission);
+		const title = `${localizeSubmission(currentPublication.fullTitle, currentPublication.locale)}`;
+
+		const {apiUrl} = useApiUrl(
+			`reviews/${submissionId}/${reviewAssignment.id}/confirmReview`,
+		);
+
+		form.action = apiUrl;
+
+		openSideModal(
+			WorkflowLogResponseModal,
+			{
+				title: title,
+				submissionId: submissionId,
+				logResponseForm: dashboardStore.componentForms.logResponseForm,
+			},
+			{onClose: finishedCallback},
+		);
+	}
+
 	return {
 		getTopActions,
 		getItemActions,
@@ -586,5 +629,6 @@ export function useReviewerManagerActions() {
 		reviewerThankReviewer,
 		reviewerRevertConsider,
 		reviewerSendReminder,
+		reviewerLogResponse,
 	};
 }
--- a/src/components/Modal/ModalManager.vue
+++ b/src/components/Modal/ModalManager.vue
@@ -70,8 +70,9 @@ import {storeToRefs} from 'pinia';
 import SideModal from '@/components/Modal/SideModal.vue';
 import LegacyAjax from '@/components/Modal/SideModalBodyLegacyAjax.vue';
 import PkpDialog from '@/components/Modal/Dialog.vue';
+import WorkflowLogResponseModal from '@/managers/ReviewerManager/modals/WorkflowLogResponseModal.vue';
 
-const GlobalModals = {LegacyAjax};
+const GlobalModals = {LegacyAjax, WorkflowLogResponseModal};
 
 const modalStore = useModalStore();
 const {
--- a/src/managers/ReviewerManager/modals/WorkflowLogResponseModal.mdx
+++ b/src/managers/ReviewerManager/modals/WorkflowLogResponseModal.mdx
@@ -1,0 +1,11 @@
+import {Meta, ArgTypes} from '@storybook/blocks';
+
+import * as WorkflowLogResponseModalStories from './WorkflowLogResponseModal.stories';
+
+<Meta of={WorkflowLogResponseModalStories} />
+
+# Log Response Modal
+
+A vue modal containing a form for and admin to log a response to a review request on behalf of the reviewer.
+
+<ArgTypes/ >
--- a/src/managers/ReviewerManager/modals/WorkflowLogResponseModal.stories.js
+++ b/src/managers/ReviewerManager/modals/WorkflowLogResponseModal.stories.js
@@ -1,0 +1,77 @@
+import WorkflowLogResponseModal from './WorkflowLogResponseModal.vue';
+import PkpButton from '@/components/Button/Button.vue';
+import {useModal} from '@/composables/useModal';
+import {within, userEvent} from '@storybook/test';
+
+export default {
+	title: 'Pages/Workflow/LogResponse',
+	component: WorkflowLogResponseModal,
+};
+
+export const Base = {
+	render: (args) => ({
+		components: {WorkflowLogResponseModal, PkpButton},
+		setup() {
+			const {openSideModal} = useModal();
+
+			function logResponse() {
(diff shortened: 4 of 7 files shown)


src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,147 @@
...
-			<div class="bg-secondary p-4">
-				<div>SubmissionId {{ submissionId }}</div>
-				<div>ReviewAssignmentId {{ reviewAssignmentId }}</div>
+		<template #description>
+			<span>{{ description }}</span>
+		</template>
+		<div class="px-8 h-full">
+			<div class="bg-secondary h-full">
+				<form class="pkpForm -pkpClearfix h-full">
+					<div class="form-flex h-full">
```
jardakotesovec, line 15: Could se use the PkpForm here? And create the form configuration on php side as we have the classes to do that there?

I know its bit of the hassle for such simple form - but its important to stay consistent on having configuration driven forms as that let us to evolve them over time.

src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,147 @@
...
+						<fieldset class="pkpFormGroup -pkpClearfix">
+							<FieldOptions
+								v-bind="logResponseForm"
+								@change="change"
+							/>
+						</fieldset>
+						<div>
+							<div class="buttonRow pkpFormPage__footer">
+								<div>
```
jardakotesovec, line 24: Was not ButtonRow working for this case?

Generally what we aim in styling is to use either tailwindcss for simple cases and when its more comples&repeating pattern we would create component - example would be SideModalLayout2Columns.

Would like to avoid using some global css classes that has been created in past - that makes very difficult to keep oriented whats used where and makes impossible to make changes. Thats also for example why we introduced vairants in Button.vue and other components rather than overriding their styling from outside - because than it makes impossible to change these components over time.

src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,147 @@
...
+			}
+		}
+
+		function change(field, label, value) {
+			acceptReview.value = value;
+		}
+
+		function submit() {
+			$.ajax({
+				url: props.url,
```
jardakotesovec, line 118: With composition API I created couple of composable helpers to handle common actions for API interactions, modals and localisation.

Please check them out in [storybook](https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/composables-usefetch--docs), in particular useModal, useFetch, useLocalise and useUrl are most common one.  We also have useForm when we need to explicitely set/get some values from the form, not sure if you would need that - thats has not been documented yet. If you think it would be useful for you let me know I will prioritize to documenting the use cases.

I will merge this improvement for useFetch soon, which I believe would be relevant for you - https://github.com/pkp/pkp-lib/issues/9977

jardakotesovec:
@MrRob100 After code improvements I will also ask you to add it to storybook, so we have that modal snapshoted, which helps with awarness that such modal do exist and it also helps with preventing some regressions over time as its screenshots are always compared to baseline.

It will require following
- create stories file, which need to have button which opens this modal via play function. That what ensures that opened modal is screenshoted
- create mdx file with simple description, so it does appear in storybook - and it automatically is able to pick up the props and show them there


One of the best examples is ReviewerSubmissionPage.stories, as that also have modals that are being opened via play function:
```
export const OpenModalAccepted = {
	play: async ({canvasElement}) => {
		// Assigns canvas to the component root element
		const canvas = within(canvasElement);
		const user = userEvent.setup();

		await user.click(canvas.getByText('Read Round 1 Review'));
	},
	decorators: [
		() => ({
                        // this needs to be adjusted based on your specific modal to ensure the whole modal gets screenshoted
			template: '<div style="height: 1600px"><story/></div>',
		}),
	],
};

```

Happy to give more guidance on mattermost for storybook if needed, there is just couple of things to watch out so the modal gets properly screenshoted.

src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,147 @@
...
+			});
+		}
+
+		function cancel() {
+	  	acceptReview.value = null;
+		  closeModal();
+		}
+
+		return {submit, change, cancel, acceptReview, logResponseForm};
+	}
```
jardakotesovec, line 136: One thing I forgot to mention. Could we also move the business logic to the WorkflowLogResponseForModalStore.js ?

Good (similar) example to this would be in roundHistoryModalStore.js which is also handling logic for modal.

Some resources why we do this and some guidance
- https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-technical-roadmap--docs#pinia-stores-35
- https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-pinia-store--docs

Good is that the syntax is completely the same as in component when using composition api. So its just matter of moving things over.

MrRob100:
I have redone this using the pkp-form component. In order to make the form footer sit on the bottom of the modal, I have had to add some styling to the Form component. I have checked and it doesn’t seem to have knock-on effects on the other forms. Also, there doesn’t seem to be support for a cancel button in the pkp-form component which is why I have left that out. I can alter it to allow for one if we like. Thoughts @jardakotesovec @Devika008

jardakotesovec (changes requested):
Here is example for storybook - https://github.com/pkp/ui-library/commit/815ea2a764bfb4eeee16217c0c36dd70346f249d

This ensures that if you open that story it opens the modal - and as result it gets automatically screenshotted in CI to ensure it looks good over time.

You can see it in storybook (cd lib/ui-library && npm run storybook).

Please just refine it a bit with some description. Otherwise I think one simple story suffice.

src/components/Form/Form.vue
```diff
@@ -672,7 +672,10 @@ export default {
 }
 
 .pkpFormPage--current {
-	display: block;
+	height: 100%;
+	display: flex;
+	flex-direction: column;
+	justify-content: space-between;
 	position: relative;
```
jardakotesovec, line 679: I think I commented on this one before. Are we sure this does not break forms with multiple groups? Like creating spaces between individual groups as well.

jardakotesovec:
@MrRob100 Thanks! Looking good, just need to do some testing. Would you be able to rebase it, so its up to date with main branches? Having difficulties to run it. There has been lots of merged lots of changes recently.

src/components/Form/Form.vue
MrRob100: @jardakotesovec I have checked and it seems to display as normal for example in the form here management/settings/context#contact.

jardakotesovec: @MrRob100 Right, but thats because these are long forms which take full page height. But if there would be short form (like yours) with multiple groups, it would equally space between individual groups.

Can you try to add margin-top: auto  to the pkpFormPage__footer - that should position button down, and not affect the group. 

Thanks!

src/components/ListPanel/doi/DoiItemVersionModal.vue
```diff
@@ -10,50 +10,42 @@
...
 			>
 				<a
-					:id="versionHeaderId"
 					:href="version.urlPublished"
 					target="_blank"
 					rel="noopener noreferrer"
 				>
 					{{ getVersionHeader(version) }}
 				</a>
-				<PkpTable :labelled-by="versionHeaderId">
```
jardakotesovec, line 20: These changes probably should not be here?

src/pages/workflow/WorkflowLogResponseForModal.stories.js
```diff
@@ -0,0 +1,77 @@
+import WorkflowLogResponseForModal from './WorkflowLogResponseForModal.vue';
```
jardakotesovec, line 1: Thank you for adding this to storybook - thats great.. it help us to be aware of this screen and because it makes screenshots it also helps to make sure it still looks correct as we change things.

src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,37 @@
...
-				<div>SubmissionId {{ submissionId }}</div>
-				<div>ReviewAssignmentId {{ reviewAssignmentId }}</div>
-			</div>
-			<PkpButton @click="submitThings">Submit Things</PkpButton>
+		<template #description>
+			<span>{{ description }}</span>
+		</template>
+		<div class="ml-8 mr-8 h-full bg-secondary">
+			<pkp-form
+				class="h-full"
```
jardakotesovec, line 14: Here we just need to npm run lint in ui-library folder (after npm install). This should happen automatically if you have dependencies installed there I think. 

It will fix the naming convention to <PkpForm>

src/pages/workflow/WorkflowLogResponseForModal.vue
```diff
@@ -1,39 +1,37 @@
...
 import SideModalBody from '@/components/Modal/SideModalBody.vue';
-import PkpButton from '@/components/Button/Button.vue';
+import PkpForm from '@/components/Form/Form.vue';
 
 const props = defineProps({
+	description: {type: String, required: true},
 	submissionId: {type: Number, required: true},
-	reviewAssignmentId: {type: Number, required: true},
 	legacyOptions: {type: Object, required: true},
+	logResponseForm: {type: Object, required: true},
```
jardakotesovec, line 33: are the legacyOptions needed actually

src/pages/workflow/workflowLogResponseForModalStore.js
```diff
@@ -0,0 +1,20 @@
...
+		const closeModal = inject('closeModal');
+		const form = props.logResponseForm;
+
+		function formSuccess() {
+			closeModal();
+		}
+
+		function updateForm(formId, data) {
+			Object.keys(data).forEach((key) => (form[key] = data[key]));
+		}
```
jardakotesovec, line 16: Its not well documented as its still in early stages, but for these cases we now have useForm composable.

```
import useForm from '@/composables/useForm'
const {set: updateForm, form} = useForm(props.logResponseForm);
```

Which provides the set function, so you don't need the updateForm function.

jardakotesovec (changes requested):
Thanks for the storybook addition. Just couple small tweaks! And also make sure it does not include changes outside of this work. Thank you!

src/components/Form/Form.vue
MrRob100: @jardakotesovec I was unable to resolve this by styling the pkpFormPage__footer. I have pushed another solution that just targets the pkpFormPage__footer that is used in the vue modal

MrRob100: @jardakotesovec Removed those css changes completely now

src/managers/ReviewerManager/useReviewerManagerActions.js
```diff
@@ -564,6 +575,30 @@ export function useReviewerManagerActions() {
...
+		const title = `${localizeSubmission(currentPublication.fullTitle, currentPublication.locale)}`;
+		form.action =
+			form.action +
+			`reviews/${submissionId}/${reviewAssignment.id}/confirmReview`;
+
+		openSideModal(WorkflowLogResponseModal, {
+			title: title,
+			submissionId: submissionId,
+			logResponseForm: dashboardStore.componentForms.logResponseForm,
+		});
```
jardakotesovec, line 599: To update data, it should be enough just to call finishedCallback, so the main store triggers data reload.

Something like this should work
```
openSideModal(WorkflowLogResponseModal, {
  title: title,
  submissionId: submissionId,
  logResponseForm: dashboardStore.componentForms.logResponseForm,
}, {onClose: finishedCallback});
```

jardakotesovec (changes requested):
Looking great! I think just make sure that the data update works after suggested change and should be ready for merging (at least the ui-part).

PR #354, opened 2024-05-21
[UI-LIBRARY][main] #9840 Remove vue3-highlightjs and add "highlightjs/vue-plugin". Use only XML language.

defstat:


src/components/CodeHighlighter/CodeHighlighter.vue
```diff
@@ -0,0 +1,64 @@
...
+			const languageModule = languageMap[language];
+			if (languageModule) {
+				hljs.registerLanguage(language, languageModule);
+			} else {
+				console.error(`Language module for ${language} is not available.`);
+			}
+		},
+	},
+};
+</script>
```
jardakotesovec, line 60: Can you please update it to use composition API? Its quite straightforward:
And please make sure the formatting is applied. I have prettier enabled in the vscode, so it formats all the time, but you can also make that happen with `npm run format`

Something like:

```html
<script>
import hljs from 'highlight.js/lib/core';
import hljsVuePlugin from '@highlightjs/vue-plugin';
// Importing the language modules
import xml from 'highlight.js/lib/languages/xml';
// Add more languages
// A map to store language modules

const highlightjs = hljsVuePlugin.component;
const languageMap = {
	xml,
	// Add more language modules imported above
};
const supportedLanguages = Object.keys(languageMap);
const props = defineProps({ 
            code: {
			type: String,
			required: true,
		},
		language: {
			type: String,
			required: true,
			validator(value) {
				if (!supportedLanguages.includes(value)) {
					console.error(
						`Unsupported language: ${value}. Supported languages are: ${supportedLanguages.join(', ')}`,
					);
					return false;
				}
				return true;
			},
		},
})

function loadLanguageModule(language) {
	const languageModule = languageMap[language];
	if (languageModule) {
		hljs.registerLanguage(language, languageModule);
	} else {
		console.error(`Language module for ${language} is not available.`);
	}
},

		
loadLanguageModule(props.language);
</script>
```

defstat: Thanks @jardakotesovec! I think I did it. I have pushed the changes

src/components/CodeHighlighter/CodeHighlighter.vue
```diff
@@ -0,0 +1,62 @@
...
+	if (languageModule) {
+		hljs.registerLanguage(language, languageModule);
+	} else {
+		console.error(`Language module for ${language} is not available.`);
+	}
+}
+
+// Validate the language prop
+function validateLanguage(language) {
+	if (!supportedLanguages.includes(language)) {
```
jardakotesovec, line 44: Validating props still should be part of defineProps - https://vuejs.org/guide/components/props.html#prop-validation

defstat: @jardakotesovec that was my initial choice also but it seems that I can't use the local variable `supportedLanguages` in the `defineProps`. Am I missing something?

src/components/CodeHighlighter/CodeHighlighter.vue
```diff
@@ -0,0 +1,62 @@
...
+			`Unsupported language: ${language}. Supported languages are: ${supportedLanguages.join(', ')}`,
+		);
+	}
+}
+
+validateLanguage(props.language);
+
+onBeforeMount(() => {
+	loadLanguageModule(props.language);
+});
```
jardakotesovec, line 55: This is probably fine, but it would call the registration always when such component is mounted. Yet it needs to happen just once.

One very simple way to achieve that is just something like this above existing <script setup> 
```html
<script>
import xml from 'highlight.js/lib/languages/xml';
import hljs from 'highlight.js/lib/core';
hljs.registerLanguage('xml', xml);
</script>
```

defstat: @jardakotesovec that code is there to load more languages if provided. But I can remove that "feature".

src/components/CodeHighlighter/CodeHighlighter.vue
jardakotesovec: @defstat And if you move the supportedLanguages from <script setup> to separate <script>, does that help?

<script>
// A map to store language modules
const languageMap = {
	xml,
	// Add more language modules imported above
};
const supportedLanguages = Object.keys(
</script>

<script setup>
defineProps...
<script>

jardakotesovec: @defstat Interestingly it seems working fine even when its in script setup - just the eslint is complaining.. will check whether its linter mistake

jardakotesovec: @defstat Alright.. so it needs to go to <script> section as illustrated above.

src/components/CodeHighlighter/CodeHighlighter.vue
jardakotesovec: @defstat but it can load something only when its imported right? The impactful thing is the import as that includes new code to the bundle. The registration is just to connect core with that language model, which is lightweight operation. If we would decide include another language - it requires both `import` and `registerLanguage` statements. Or do I missing something?

I don't mind if the registerLanguage is called lazily as longs as its done just once.

PR #355, opened 2024-05-22
pkp/pkp-lib#9977 add isSuccess computed property to useFetch

jardakotesovec:


src/composables/useFetch.js
```diff
@@ -91,12 +92,15 @@ export function useFetch(url, options = {}) {
 		}
 
 		isLoading.value = true;
+		isSuccess.value = null;
 		try {
 			const result = await ofetchInstance(unref(url), opts);
 			data.value = result;
 			validationError.value = null;
+			isSuccess.value = true;
```
ewhanson, line 100: In the issue you mentioned this will apply when a 200 status is returned. Does this cover any 200-level status code or only the actual 200 HTTP OK status code?

jardakotesovec: useFetch is using ofetch and thats using fetch and that follows fetch behaviour on this, which is that 2xx is considered 'ok'. Therefore when I say 200 it basically means 2xx

ewhanson: Perfect!

PR #363, opened 2024-06-17
pkp/pkp-lib#10033 Add new dropdown menu component

blesildaramirez:


src/components/DropdownActions/DropdownActions.vue
```diff
@@ -0,0 +1,117 @@
...
+				enter-active-class="transition ease-out duration-100"
+				enter-from-class="transform opacity-0 scale-95"
+				enter-to-class="transform opacity-100 scale-100"
+				leave-active-class="transition ease-in duration-75"
+				leave-from-class="transform opacity-100 scale-100"
+				leave-to-class="transform opacity-0 scale-95"
+			>
+				<MenuItems
+					class="dropdown-shadow absolute z-10 w-max border border-light bg-secondary focus:outline-none"
+					:class="
```
jardakotesovec, line 42: What about using 'shadow' classes that we have configured in tailwind? Its slightly smaller shadow - but if we would want to introduce additional shadow we would add it as variant to tailwind..

src/components/DropdownActions/DropdownActions.stories.js
```diff
@@ -0,0 +1,169 @@
...
+				defaultValue: {summary: ''},
+			},
+			description:
+				"The text label for the button. This is optional. If `isEllipsis` is `false` and `name` is provided, it will be used as the button's label.",
+		},
+		position: {
+			control: {type: 'select'},
+			options: ['left', 'right'],
+			description:
+				'Determines where to show the dropdown button. Options include `left` and `right`.',
```
jardakotesovec, line 37: Aiming to have the documentation collocated with the component. Storybook si pretty good in picking up jsdoc.. So at least description and options should go there..   Not sure if will be possible to make it 'type Select' just from jsdoc, but please check if thats possible.

You can see for example in Button.vue the comments which are than displayed in storybook automatically. There has been some improvements in latest storybook - so not sure whether there is more than description that could be conveyed..

src/components/DropdownActions/DropdownActions.vue
```diff
@@ -0,0 +1,117 @@
...
+				>
+					<MenuItem v-for="(action, i) in actions" :key="i" v-slot="{active}">
+						<div class="min-w-[96px]">
+							<PkpButton
+								v-if="action.label"
+								element="a"
+								:href="action.url"
+								:icon="action.icon"
+								:is-active="active"
+								:is-warnable="action.isWarnable"
```
jardakotesovec, line 56: Actions rarely will be links - mostly it will be actions that just needs to be emitted and handled by the parent component.

src/components/DropdownActions/DropdownActions.vue
```diff
@@ -0,0 +1,117 @@
...
+	},
+	position: {
+		type: String,
+		default: 'right',
+	},
+});
+
+const computedAriaLabel = computed(() => {
+	return props.ariaLabel || props.name || 'More actions';
+});
```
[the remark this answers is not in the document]
blesildaramirez: As discussed offline, we will require the "name" prop and rename it to "label" to make it consistent to other components. I updated this now.

src/components/DropdownActions/DropdownActions.vue
blesildaramirez: This is now addressed, please see latest updates on the PR :)

src/components/DropdownActions/DropdownActions.vue
blesildaramirez: I reverted introducing a new class and instead used the existing "shadow" class.

PR #366, opened 2024-06-25
I9992

jardakotesovec:


src/components/FileAttacher/AttacherModal.vue
```diff
@@ -0,0 +1,40 @@
...
+		<template #title>
+			{{ title }}
+		</template>
+		<SideModalLayoutBasic>
+			<component
+				:is="Components[currentAttacher.component] || currentAttacher.component"
+				v-if="currentAttacher"
+				v-bind="currentAttacher"
+				@selected:files="(...args) => emit('attachFiles', ...args)"
+				@cancel="(...args) => emit('cancel', ...args)"
```
ewhanson, line 12: Just FYI: my IDE was labeling the `@cancel` event as deprecated, but I couldn't find anything more about it.

src/pages/statsContext/ContextDownloadReportModal.vue
```diff
@@ -0,0 +1,53 @@
...
+		</template>
+		<SideModalLayoutBasic>
+			<p>{{ t('stats.context.downloadReport.description') }}</p>
+			<table class="pkpTable pkpStats__reportParams">
+				<tr class="pkpTable__row">
+					<th>{{ t('stats.dateRange') }}</th>
+					<td>{{ dateRangeDescription }}</td>
+				</tr>
+			</table>
+			<action-panel class="pkpStats__reportAction">
```
ewhanson, line 14: I noticed here `<action-panel>` and `<pkp-button>` are used whereas other places (I noticed in `PublicationsDowloadReportModal` but wasn't comprehensive in checking everywhere) `<PkpButton>` and `<ActionPanel>` are used. I know both will work, but wondering if we've standardized on a "house style" for this and if so, if it could be encapsulated in an eslint rule or something?

src/components/ListPanel/doi/DoiFailedActionDialogBody.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+	<p>{{ t('manager.dois.update.partialFailure') }}</p>
+	<ul>
+		<li v-for="errorMessage in failedDoiActions" :key="errorMessage.index">
```
ewhanson, line 4: The `<li>` items have `list-style: none;` applied to them (which wasn't the case previously).

DOI failed updates in OJS 3.4

From OJS 3.4 ☝️


DOI failed updates modal in this PR

From this PR ☝️

src/components/ListPanel/doi/DoiItemVersionModal.vue
```diff
@@ -0,0 +1,109 @@
...
+								type="text"
+								:readonly="!(isEditingDois && !isSaving)"
+							/>
+						</table-cell>
+					</template>
+				</pkp-table>
+			</div>
+
+			<div class="doiListItem__versionContainer--actionsBar">
+				<spinner v-if="isSaving" />
```
ewhanson, line 47: It seems to be working as intended, but neither `<spinner>` nor `<pkp-button>` on the next line have been imported. Not sure if it's an issue, but wanted to flag it for you.

src/components/ListPanel/highlights/HighlightsListPanel.vue
```diff
@@ -421,3 +423,4 @@ export default {
 	}
 }
 </style>
+./HighlightsEditModal.vue
```
ewhanson, line 426: I'm guessing this probably isn't supposed to be here.

src/pages/submissionWizard/ReconfigureSubmissionModal.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+	<SideModalBody>
+		<template #title>
+			{{ t('submission.wizard.changeSubmission') }}
```
ewhanson, line 4: Makes use of `t` but doesn't have the corresponding `const {t} = useLocalize();` in the script below.

src/components/Modal/SideModal.mdx
```diff
@@ -13,13 +13,15 @@ import * as SideModalStories from './SideModal.stories.js';
...
 
-Side modal has been introduced with new submission listing and will gradually replace both legacy jQuery modal and Vue.js central modal.
-
-To correctly handle accessibility (title, description) and focus management - [headless-ui](https://headlessui.com) library is used.
+Side modal has been introduced in 3.5 all existing modals, including jquery legacy ones are now rendered as SideModal.
 
 ## Usage
 
-We recommend to define modals as individual component files, rather than inline them, that ensures that the `setup` function and any other component life cycle event are triggered as modal is opened/closed. Therefore its easier to control when for example to fetch data from API. We might introduce option to define inline modals for basic use cases in future
+Its important to create modals as separate files, which works best with our [useModal](../?path=/docs/composables-usemodal--docs#opensidemodal) composable. Also having them defined as individual components ensures that the `setup` function and any other component life cycle event are triggered as modal is opened/closed. Therefore its easier to control when for example to fetch data from API.
```
ewhanson, line 20: Both its should be `It's`. `Therefore it's easier` could also be `Therefore, it's easier`, but that's more of a style preference.

src/composables/useModal.mdx
```diff
@@ -1,14 +1,10 @@
...
-import * as useModalStories from './useModal.stories.js';
-
-<Meta of={useModalStories} />
-
 # useModal
 
 ## openDialog
 
-Dialogs provide a quick way to show a simple confirmation prompt. Import `useModal` composable and use the openDialog() method to create a dialog.
+Dialog purpose is to display simple feedback like success and error messages. Or request confirmation for example before deleting some content. Import `useModal` composable and use the openDialog() method to create a dialog. Check [Dialog](../?path=/docs/components-dialog--docs) component for more details on available props.
```
ewhanson, line 7: Another minor wording suggestion: instead of `Dialog purpose`, maybe `The Dialog component's purpose`.

src/pages/statsContext/ContextDownloadReportModal.vue
jardakotesovec: @ewhanson I am almost always aiming for vue defaults. Which in case is to use ActionPanel in vue SFC files and only to use action-panel when its defined in 'html', which in our cases would be in smarty templates that we are moving away from.  Having eslint rule for this is good idea, will check that out.

src/components/ListPanel/doi/DoiItemVersionModal.vue
jardakotesovec: @ewhanson Thanks - its because most component are registered globally. But I think still there is value to import dependencies always so its clear on which components it depends..  for example in storybook its important.

I wish this would be caught by tooling like eslint as well.. maybe its possible, will check.

src/composables/useModal.mdx
jardakotesovec: Thank you, wording suggestions are much appreciated, not my strong suite :-)

PR #372, opened 2024-06-27
pkp/pkp-lib##10056 Prevent dropdown from taking whitespace input in autosuggest

Hafsa-Naeem:
Fixes [#10056](https://github.com/pkp/pkp-lib/issues/10056)

### DIFF
--- a/src/components/Form/fields/FieldBaseAutosuggest.vue
+++ b/src/components/Form/fields/FieldBaseAutosuggest.vue
@@ -119,7 +119,7 @@
 						@blur="() => (isFocused = false)"
 					/>
 					<ComboboxOptions
-						v-if="suggestions.length || (allowCustom && inputValue.trim().length)"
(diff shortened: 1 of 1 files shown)


src/components/Form/fields/FieldBaseAutosuggest.vue
```diff
@@ -114,12 +114,12 @@
 						ref="autosuggestInput"
 						class="pkpAutosuggest__input"
 						v-bind="inputProps"
-						@change="inputValue = $event.target.value"
+						@change="inputValue = $event.target.value.trim()"
 						@focus="() => (isFocused = true)"
```
jardakotesovec, line 118: Looking good, just wondering - is not triming here enough? I would expect that the inputValue always will be trimmed as result and its not needed to trim it on other places again, or do I miss some scenario?

PR #373, opened 2024-07-01
pkp/pkp-lib#8248 COUNTER R5 TSV reports

bozana:
s. https://github.com/pkp/pkp-lib/issues/9666


src/components/Form/counter/CounterReportForm.vue
```diff
@@ -0,0 +1,137 @@
+<script>
+import PkpForm from '@/components/Form/Form.vue';
+
+export default {
+	name: 'CounterReportForm',
+	extends: PkpForm,
+	components: {
```
jardakotesovec, line 7: Goal here will be to stop extending and overriding form methods. Form component should provide good enough API via props and emits to achieve these custom submits and I suspect it already can. And if not we can improve it.

I see in the code that Form already has option to only emit success action, when the action === 'emit', thats what might be good enough to handle the submitting outside of the form?

Additionally we have now composables/useForm, which unfortunately is not documented yet as its still evolving quite a bit, but its purpose is to make easier to programatically change or retrieve form state.

As very basic use can be just for maintaining form state - since we don't inherit from Page.vue, which has the `set` method, you should be able to do something like

const {form, set} = useForm(props.statsForm)

and use the `form` and `set` on the <PkpForm> component.

Please check it out - and feel free to follow-up so we can improve things.

src/components/ListPanel/counter/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+	<div class="counterReportsListPanel">
+		<slot>
+			<list-panel :items="items">
+				<pkp-header>
```
jardakotesovec, line 5: We don't have mature enough replacement for ListPanel, so nothing to change on that front.

Only things that would make sense to leverage apart from compositionAPI would be to useUrl and useFetch for the api communication. These composables will abstract many things for you, like handling headers, csrf token or displaying error if backend fails.

 You should be able to find good examples both in storybook and in existing code.

bozana:
Hi @jardakotesovec, I believe this is ready for your next review :-)
I think I have considered all suggestions from the first review round.
I have integrated your changes for the customSubmit to work, and slightly adapted it in the commit [adapt form](https://github.com/pkp/ui-library/pull/373/commits/3dbf5d24c918ff8378509e5b34a312a6f52595fa).
And finally, why are the tests failing i.e. how should I proceed regarding them?
Thanks a lot!

src/components/ListPanel/counter/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,193 @@
...
+</script>
+
+<style lang="less">
+@import '../../../styles/_import';
+
+.counterReportsListPanel {
+	.listPanel__itemTitle {
+		font-weight: @normal;
+	}
+}
```
jardakotesovec, line 192: Generally trying to minimise cases where we override component styles from outside, as it makes difficult to evolve the component on its own when consumer are relying on such implementation details.

But in this case you can easily adjust the styling like this:
```
					<span :id="item.Report_ID" class="text-lg-normal">
						{{ item.Report_Name }} ({{ item.Report_ID }})
					</span>
```

src/components/ListPanel/counter/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,193 @@
...
+		PkpButton,
+		Spinner,
+		ListPanel,
+		PkpHeader,
+	},
+	mixins: [fetch, ajaxError],
+	props: {
+		/** A localized string for the button to edit a report. */
+		editCounterReportLabel: {
+			type: String,
```
jardakotesovec, line 49: We don't need these anymore in 3.5.  Just use 

this.t in options api
or in composition api - import {useLocalize} from '@/composables/useLocalize' and const {t} = useLocalize()

And it will automatically get that translation down to the client without explicitly passing that from backend.

src/components/ListPanel/counter/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,193 @@
...
+			required: true,
+		},
+		/** The Form to edit a report. */
+		form: {
+			type: Object,
+			required: true,
+		},
+		/** A unique id for this component. */
+		id: {
+			type: String,
```
jardakotesovec, line 59: I don't think this is needed anymore.. but does not hurt..

src/components/ListPanel/counter/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,193 @@
...
+		},
+
+		/**
+		 * Open the modal to edit an item
+		 *
+		 * @param {Number} id
+		 */
+		openEditModal(id) {
+			const report = this.items.find((report) => report.Report_ID === id);
+			if (!report) {
```
jardakotesovec, line 132: Can this actually happen?

src/pages/counter/CounterReportsPage.vue
```diff
@@ -0,0 +1,31 @@
...
+	</Panel>
+</template>
+
+<script setup>
+import Panel from '@/components/Panel/Panel.vue';
+import PanelSection from '@/components/Panel/PanelSection.vue';
+import CounterReportsListPanel from '@/components/ListPanel/counter/CounterReportsListPanel.vue';
+
+defineProps({
+	counterReportsListPanel: {type: Array, required: true},
```
jardakotesovec, line 26: Is it really Array? I would expect this to be Object?

jardakotesovec (changes requested):
Nothing major really..  but suggested couple of improvements.

Really appreciate willingness to familiarise with the composition api even its not completely necessary for this particular task. There is some getting used to - but it really gives lots of power for more complex UIs.

src/components/ListPanel/counter/CounterReportsEditModal.vue
```diff
@@ -0,0 +1,109 @@
...
+		await fetch();
+
+		if (
+			validationError.value &&
+			Object.prototype.hasOwnProperty.call(validationError.value, 'Code')
+		) {
+			// COUNTER speific errors should actually not occur
+			// because of the form/user input validation
+			// but consider them for any case as well.
+			pkp.eventBus.$emit(
```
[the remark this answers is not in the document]
bozana: The error comes from the API itself that also checks a few things. The form is COUNTER specific -- it is a JSON, for example:
`{"Code":1030,"Severity":"Fatal","Message":"Insufficient Information to Process Request","Data":"Missing customer_id."}`.
An example where it happened: https://github.com/pkp/pkp-lib/issues/10350 -- I was treating the UI form inputs wrongly, so API error occurred.

src/components/ListPanel/counter/CounterReportsListPanel.vue
bozana: I will then leave it for now...

src/components/ListPanel/counter/CounterReportsListPanel.vue
bozana: Theoretically it cannot happen :-)
Would it be possible that someone manipulates the list panel and edit function?
However, I will remove it :-)

src/components/ListPanel/counter/CounterReportsEditModal.vue
```diff
@@ -0,0 +1,109 @@
+<template>
```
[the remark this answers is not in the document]
bozana: Hi @jardakotesovec, I am not sure I understand this correctly.
We now have the page src/pages/counter/CounterReportsPage.vue. 
Should this file src/components/ListPanel/counter/CounterReportsEditModal.vue be copied to src/pages/counter/CounterReportsEditModal.vue?
:thinking:

bozana:
Hi @jardakotesovec, I believe I have considered all your comments -- only one question is left -- so it is ready for the next review round :-)
Thanks a lot!

src/components/ListPanel/counter/CounterReportsEditModal.vue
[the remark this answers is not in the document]
bozana: Aaaa... OK, will do... :-)

jardakotesovec (commented):
Just couple small suggestions, nothing major.

src/pages/counter/components/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,102 @@
...
+
+	const {apiUrl} = useApiUrl(`stats/sushi/reports`);
+	const {data, isSuccess, fetch} = useFetch(apiUrl, {
+		method: 'GET',
+	});
+	await fetch();
+
+	if (isSuccess) {
+		items.value = data.value;
+	}
```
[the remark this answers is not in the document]
bozana: But I could restructure it all a little bit so that the function getItems is not there...

src/pages/counter/CounterReportsPage.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+	<h1 class="app__pageHeading">
+		{{ title }}
+	</h1>
+	<p v-html="description" />
+	<Notification
+		v-if="usageNotPossible"
+		type="warning"
+		class="pkpNotification--backendPage__header"
```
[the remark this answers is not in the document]
bozana: I hope I understood this correctly :-)

bozana:
Hi @jardakotesovec, I think I considered everything, could you please take another look?
Thanks a lot!

src/pages/counter/components/CounterReportsListPanel.vue
```diff
@@ -0,0 +1,87 @@
...
+	data,
+	fetch,
+	isLoading: isLoadingItems,
+} = useFetch(apiUrl, {
+	method: 'GET',
+});
+
+onMounted(async () => {
+	await fetch();
+});
```
jardakotesovec, line 62: Just FYI, you don't need to trigger fetch inside onMounted. The whole setup function gets called when the component is about to be created (before onMounted). So I usually call fetch directly inside the setup function:
```
const {
	data,
	fetch,
	isLoading: isLoadingItems,
} = useFetch(apiUrl, {
	method: 'GET',
});

fetch()
```

But it does not make  real difference.

jardakotesovec (approved):
Thank you for going along with composition api and other new things!

src/pages/counter/components/CounterReportsListPanel.vue
bozana: I have changed it... Thanks!

jardakotesovec (changes requested):

PR #374, opened 2024-07-02
[TEST ONLY] fix ui issues with tailwind css

ipula:


src/pages/acceptInvitation/AcceptInvitationPage.vue
```diff
@@ -1,12 +1,7 @@
...
-		<AcceptInvitationHeader
-			:page-title="store.pageTitle"
-			:page-title-description="store.pageTitleDescription"
-		/>
+	<div v-if="store.steps.length">
 		<Steps
-			v-if="store.steps.length"
-			class="userInvitation__steps"
+			class="border-[1px] border-b-0 border-solid border-[#ddd]"
 			:current="store.currentStep.id"
```
jardakotesovec, line 5: I would recommend `border-x border-t border-light`.  Using particular values is the last resort. In storybook you can see what border colors we have as options (in this case its border-light).

Class `border` gives you 1px border.  No need to explicitely define 1px.  If you want to apply it on left, right and top - you can use the border-x and border-t variants. I would be careful about overriding styles like with border-b-0. That might lead to unexpected results because it depends on the order of classes in the sheet and you can't really affect it by the order in the class attribute.

src/pages/acceptInvitation/AcceptInvitationPage.vue
```diff
@@ -21,34 +16,46 @@
...
-				</Panel>
+				<div class="border-b-[1px] border-solid border-[#ddd]">
+					<div class="flex border-b-[1px] border-solid border-[#ddd] p-4">
+						<div class="min-w-px flex-1">
+							<h2 class="m-0 text-xl-bold text-heading">
+								{{ store.stepTitle }}
+							</h2>
+							<p class="text-red font-bold">{{ store.errors.error }}</p>
+							<p class="mb-0 ml-0 mr-0 mt-1 text-lg-normal leading-5">
+								{{ step.description }}
```
jardakotesovec, line 27: Are you sure you need the zero margins? Would not be that by default?

We already define line height within the text-lg-normal. I think that should give you good default, without changing it?

src/pages/acceptInvitation/AcceptInvitationPage.vue
```diff
@@ -21,34 +16,46 @@
...
-					</PanelSection>
-				</Panel>
+				<div class="border-b-[1px] border-solid border-[#ddd]">
+					<div class="flex border-b-[1px] border-solid border-[#ddd] p-4">
+						<div class="min-w-px flex-1">
+							<h2 class="m-0 text-xl-bold text-heading">
+								{{ store.stepTitle }}
+							</h2>
+							<p class="text-red font-bold">{{ store.errors.error }}</p>
+							<p class="mb-0 ml-0 mr-0 mt-1 text-lg-normal leading-5">
```
jardakotesovec, line 26: This is the error you are styling here, right?  Available colors are also in storybook - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-style-colors--docs 

In this case it would be text-negative
 
Screenshot 2024-07-16 at 13 16 17

src/pages/acceptInvitation/AcceptInvitationPage.vue
```diff
@@ -21,34 +16,46 @@
...
+								{{ step.description }}
+							</p>
+						</div>
+					</div>
+					<div v-for="section in step.sections" :key="section.id" class="flex">
+						<div class="min-w-px flex-1">
+							<template v-if="step.type === 'review'">
```
jardakotesovec, line 33: Just curious - is the min-w-px needed here?

src/pages/acceptInvitation/AcceptInvitationPage.vue
```diff
@@ -21,34 +16,46 @@
...
+						</div>
+					</div>
+				</div>
 			</Step>
 		</Steps>
-		<ButtonRow class="panel panel--wide userInvitationForm__footer">
-			<PkpButton @click="store.cancel">Cancel</PkpButton>
+		<ButtonRow
+			class="panel panel--wide flex items-center justify-end border-[1px] border-t-0 border-solid border-[#ddd] p-8"
+		>
```
jardakotesovec, line 53: do we need the panel here? ButtonRow should handle its own styling - if you want to make border around it - better would be to add parent div and apply it there rather than trying to override ButtonRow styling.

I did not test this.. so might be wrong..  But either ButtonRow is doing what you need and you don't need apply styling there. Or if not - you could just position it and style it without ButtonRow..

Please follow-up if needed.

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -1,82 +1,68 @@
...
+					:heading="t('user.password')"
+					:value="store.acceptinvitationPayload.password"
+				></FormDisplayItemBasic>
 			</div>
 		</div>
-		<div class="acceptInvitation__reviewPanel">
-			<div class="acceptInvitation__reviewPanel__header">
-				<h3>User Details</h3>
+		<div class="border-t border-solid border-[#ddd] p-8">
+			<div class="flex items-center p-4">
```
jardakotesovec, line 21: This needs to be updated on couple of places to use border-light

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -1,82 +1,68 @@
...
+				<h3>{{ t('acceptInvitation.review.userDetails') }}</h3>
 				<PkpButton
 					v-if="store.userId === null"
-					class="acceptInvitation__reviewPanel__edit"
+					class="ms-auto"
 					@click="store.openStep('userDetails')"
 				>
```
jardakotesovec, line 27: This is ok, but generally trying to avoid applying styling on existing components. Therefore I usually wrap it. But in this case I think good fit instead of the automatic margin is `justify-between` class coming from flexbox.

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
```diff
@@ -6,48 +6,56 @@
...
+				:heading="t('user.orcid')"
+				:value="store.invitationPayload.orcid"
+			></FormDisplayItemBasic>
+
+			<Icon
+				v-if="store.invitationPayload.orcidValidation"
+				icon="orcid"
+				:inline="true"
+			/>
+
```
jardakotesovec, line 28: Does this work in terms of the formatting? Or do we need to come up with alternative strategy to sneak in the orcid icon there?

src/pages/userInvitation/UserInvitationSearchFormStep.vue
```diff
@@ -1,31 +1,44 @@
...
-	<FieldText
-		:label="t('user.orcid')"
-		:value="fields.orcid"
-		name="orcid"
-		size="large"
-		@change="updateField"
-	/>
+	<div class="p-8">
+		<div class="my-4">
+			<FieldText
```
jardakotesovec, line 4: This is fine - nice alternative if you have lots of items where you want to have equal spacing is to use space-y-4 - https://tailwindcss.com/docs/space#add-vertical-space-between-children

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
ipula: yes for this we need alternative strategy .  There two icons to show here 

1. exclamation-triangle icon with orcid not verified message
2. orcid icon

PR #379, opened 2024-07-05
pkp/pkp-lib#10170 Fixed synchronization of the reviewer container

jonasraoni:


jonasraoni:
@jardakotesovec I'm not following the updates at the `ui-library` closely, so please review this PR. I've basically just added the `@set`.

src/components/Container/AdvancedSearchReviewerContainer.vue
```diff
@@ -5,34 +5,37 @@
...
-							class="author_row"
-							v-for="(affiliations, author) in displayedItems"
-							:key="author"
+						class="author_row"
+						v-for="(affiliations, author) in displayedItems"
+						:key="author"
 					>
 						<strong>{{ author }}</strong>
-						<span v-if="affiliations"> - </span>
+						<span v-if="affiliations">-</span>
```
jonasraoni, line 13: Hmm, the formatter dropped the space, I think I'll add a `&nbsp;` unless there's something better (CSS class with padding?!).

src/components/Container/AdvancedSearchReviewerContainer.vue
```diff
@@ -5,34 +5,37 @@
...
 						{{ labels.showLess }}
 					</span>
 				</div>
 			</div>
 		</div>
-		<select-reviewer-list-panel v-bind="components.selectReviewer" />
+		<select-reviewer-list-panel
+			v-bind="components.selectReviewer"
+			@set="updateReviewerList"
+		/>
```
jardakotesovec, line 38: This would work fine - for consistency with other containers - I would suggest to extend from Container component, which gives you the `set` function, so you don't have to reimplement it here and just use it with @set="set"

src/components/Container/AdvancedSearchReviewerContainer.vue
jardakotesovec: Easiest in this case would be just use `{{ ' - ' }}`   :-)

jardakotesovec:
@jonasraoni Thanks thats a good catch, added some comments.

src/components/Container/AdvancedSearchReviewerContainer.vue
jonasraoni: I'll revert this modification for now, just to keep things working (there's a circular reference).

jardakotesovec: @jonasraoni I am 93% sure that AdvancedSearchReviewerContainer can be removed from Container. I don't see reason why it would be importer there - that might have been done by mistake. Than the extending should work as expected.

jonasraoni: The UI worked fine after I removed it from there, but I didn't want to create side-effects and went with the simplest fix. As there's no reason for it to be there, I'll remove 😁

PR #382, opened 2024-07-15
[TEST ONLY]Invitation table view

ipula:


src/pages/invitation/InvitationPage.vue
```diff
@@ -0,0 +1,103 @@
+<template>
```
jardakotesovec, line 1: Was bit confused by the name of the PR - this is not really invitation landing page.  Its the new component that will be added to the User&Role page, right?

For that best fit is to implement it as self-contained component. I decided to create 'managers' folder for these, as I am also creating similar components to for example manage files. Which has lots of similarities to what you are creating.

Good thing is that these managers have very much the same structure as page. So they have main component and store. And additionally helper components as needed and also additional composables as needed so the store does not become too bloated. Main store should really just provide good quick overview how it all works together.

So I would recommend to transform this to such manager. You can check out that managers/FileManager for inspiration.

I will add some additional comments about things that could be improved along the way.

Screenshot 2024-07-18 at 12 20 54

src/pages/invitation/InvitationPage.vue
```diff
@@ -0,0 +1,103 @@
...
+	<div
+		v-if="!isInvitationLoading"
+		class="m-0 flex items-center pb-2 pt-2 text-3xl-normal leading-8"
+	>
```
jardakotesovec, line 5: Lets avoid usin the leading classes, as the line height is already defined as part of the text-3xl-normal class.

src/pages/invitation/InvitationPage.vue
```diff
@@ -0,0 +1,103 @@
...
+		</TableHeader>
+		<TableBody>
+			<TableRow v-for="(row, index) in store.invitations?.data" :key="index">
+				<TableCell>
+					{{
+						row.user?.fullName
+							? row.user.fullName
+							: localize(row.payload.givenName) +
+								' ' +
+								localize(row.payload.familyName)
```
jardakotesovec, line 30: For more complex logic is nicer to create dedicated Cell component that includes that logic in form of computed for example, rather all of that within template

src/pages/invitation/InvitationPage.vue
```diff
@@ -0,0 +1,103 @@
...
+					}}
+					<Icon icon="orcid" :inline="true" />
+				</TableCell>
+				<TableCell>
+					{{ row.email }}
+				</TableCell>
+				<TableCell>
+					<template
+						v-for="(userGroups, i) in JSON.parse(row.payload.roles)"
+						:key="i"
```
jardakotesovec, line 40: this is also good candidate for seperate cell component. Also the whole payload should be already parsed. Not sure why the roles should be send as string? In other words JSON.parse should not be necessary here.

src/pages/invitation/InvitationPage.vue
```diff
@@ -0,0 +1,103 @@
...
+					{{
+						row.user?.affiliation
+							? localize(row.user?.affiliation)
+							: localize(row?.payload.affiliation)
+					}}
+				</TableCell>
+				<TableCell>
+					<Dropdown label="...">
+						<ul>
+							<li class="border-b border-solid border-[#ddd]">
```
jardakotesovec, line 59: We already discussed this one - this would be handled by the DropdownActions component

src/pages/invitation/InvitationPageStore.js
```diff
@@ -0,0 +1,71 @@
...
+			isLoading: isInvitationLoading,
+			fetch: fetchInvitation,
+		} = useFetch(getInvitationApiUrl, {
+			query: {page: currentPage},
+		});
+		watch(
+			[currentPage],
+			async () => {
+				announce(t('common.loading'));
+
```
jardakotesovec, line 53: This is good idea, but we don't really have well established pattern for this. Will get to it sometime later on, so feel free to remove it. It will be probably more integrated to the fetching functions, will see.

src/pages/invitation/InvitationPageStore.js
```diff
@@ -0,0 +1,71 @@
...
+		function sendInvitation() {
+			window.location = sendInvitationPageUrl.value;
+		}
+		const {
+			data: invitations,
+			isLoading: isInvitationLoading,
+			fetch: fetchInvitation,
+		} = useFetch(getInvitationApiUrl, {
+			query: {page: currentPage},
```
jardakotesovec, line 47: useFetchPaginated should be better fit to calculate paginations - could we use that instead? Just for consistency so the pagination API interactions would be always handled by useFetchPaginated.

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,141 @@
...
+			},
+			{immediate: true},
+		);
+
+		/**
+		 * create user full name
+		 * @param String givenName
+		 * @param String familyName
+		 * @returns String
+		 */
```
withanage, line 72: /**
		 * create user full name
		 * @param String givenName
		 * @param String familyName
		 * @returns String
		 */

I think this  has to be get User full name

FYI @ipula

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,141 @@
...
+		function getFullName(givenName, familyName) {
+			return localize(givenName) + ' ' + localize(familyName);
+		}
+
+		function getAllInvitedRoles(userGroups) {
```
withanage, line 77: FYI @ipula 

add a comment to this function
may be "get all invited roles"

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		return {
+			invitationCount,
+			setCurrentPage,
+			sendInvitation,
+			currentPage,
+			invitationsPagination,
+			pageUrl,
+			// invitation table data
```
withanage, line 134: remove this comment
FYI @ipula

src/managers/InvitationManager/mocks/invitationMock.js
```diff
@@ -0,0 +1,97 @@
...
+						fr_CA: 'Directeur-trice de la revue',
+					},
+					masthead: true,
+					dateStart: '2024-09-10',
+					dateEnd: null,
+				},
+			],
+			userGroupsToRemove: [],
+			username: null,
+			sendEmailAddress: 'ipula7@mailinator.com',
```
withanage, line 35: i would use a comon name like journalmanager@mailinator.com

FYI @ipula

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -47,26 +50,62 @@ export const useAcceptInvitationPageStore = defineComponentStore(
...
+					'userGroupsToRemove',
+					data.value.userGroupsToRemove,
+					true,
+				);
+				updateAcceptInvitationPayload(
+					'userCountry',
+					data.value.country,
+					userId.value ? true : false,
+				);
+				// check user already exists or new user , for the new user set initial username and password for the validations
```
withanage, line 98: @ipula  i would remove this comment

src/pages/acceptInvitation/AcceptInvitationUserRoles.vue
```diff
@@ -22,6 +23,22 @@
...
+			<TableRow v-for="(row, index) in userGroupsToAdd" :key="index">
+				<TableCell :is-row-header="true">
+					{{ localize(row.userGroupName) }}
+				</TableCell>
+				<TableCell>{{ row.dateStart }}</TableCell>
+				<TableCell>{{ row.dateEnd ? row.dateEnd : '---' }}</TableCell>
+				<TableCell>
+					{{
+						row.masthead
+							? 'Appear on the masthead'
```
withanage, line 37: @ipula  can we make this come from a tempalte variable ?

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -185,17 +191,67 @@ export const useUserInvitationPageStore = defineComponentStore(
...
 				});
 			}
-
 			return error;
 		});
 
+		/**
+		 * if user already in the system userId will be sent
+		 * if user already not in the system email will be sent
+		 */
```
withanage, line 223: @ipula i think this comment should be deleted,

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -295,7 +370,9 @@ export const useUserInvitationPageStore = defineComponentStore(
 			stepTitle,
 			openStep,
 			currentStepErrorsPerSection,
-
+			isSubmitting,
+			invitationMode,
+			updateWithSelectedUserGroups,
 			//page values
```
withanage, line 376: @ipula  remove comment

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -32,55 +32,92 @@
 					</PkpButton>
 				</TableCell>
 			</TableRow>
-			<TableRow v-for="(row, index) in allUserGroups" :key="index">
+			<TableRow v-for="(row, index) in allUserGroupsToaAdd" :key="index">
```
withanage, line 35: @ipula  here is a typo.

 allUserGroupsToaAdd

after To  'a'

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -139,21 +191,30 @@ const availableUserGroups = computed(() => {
 function addUserGroup() {
```
withanage, line 191: @ipula  commented neede.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -111,21 +148,36 @@ import PkpButton from '@/components/Button/Button.vue';
...
+const allUserGroupsToaAdd = computed(
+	() => store.invitationPayload.userGroupsToAdd,
+);
 
-const props = defineProps({
-	userGroups: {type: Object, required: true},
-});
+store.updateWithSelectedUserGroups(props.userGroups);
 
 function updateUserGroup(index, fieldName, newValue) {
```
withanage, line 174: @ipula  comment needed.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -111,21 +148,36 @@ import PkpButton from '@/components/Button/Button.vue';
...
 	const userGroupsUpdate = [...store.invitationPayload.userGroupsToAdd];
```
withanage, line 175: @ipula  comment needed.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -139,21 +191,30 @@ const availableUserGroups = computed(() => {
...
 		masthead: null,
 	});
-	store.updatePayload('userGroupsToAdd', userGroupsUpdate);
+	oldUserGroupsToAdd.value.push({
+		visible: false,
+	});
+	store.updatePayload('userGroupsToAdd', userGroupsUpdate, false);
 }
 
 function removeUserGroup(userGroup, index) {
```
withanage, line 204: @ipula  comment needed.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -139,21 +191,30 @@ const availableUserGroups = computed(() => {
...
-	store.updatePayload('userGroupsToRemove', userGroupsToRemove);
+	let userGroupsToRemove = [];
+	if (store.invitationPayload.userGroupsToRemove) {
+		userGroupsToRemove = [...store.invitationPayload.userGroupsToRemove];
+	}
+	userGroupsToRemove.push({userGroupId: userGroup.id});
+	store.updatePayload('userGroupsToRemove', userGroupsToRemove, false);
 }
 
+function modifyUserGroup(index) {
```
withanage, line 214: @ipula  comment needed.

withanage (changes requested):
@ipula  I  have reviewed some of the general coding things.

public/globals.js
```diff
@@ -560,6 +560,17 @@ window.pkp = {
 		'acceptInvitation.modal.button':'View All Submissions',
 		'acceptInvitation.privacyStatement.btn':'Privacy Statement',
 		'acceptInvitation.privacyStatement.label':'Yes, I agree to have my data collected and stored according to the',
+		'invitation.cancel': 'Cancel Invite',
+		'invitation.inviteToRole.btn': 'Invite to a role',
+		'invitation.header': 'Invitation',
+		'invitation.tableHeader.name': 'Name',
+		'invitation.searchForm.emptyError': 'At least provide one search criteria.',
+		'invitation.cancelInvite.actionName':'Cancel Invite',
```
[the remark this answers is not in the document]
ipula: In the mockups its says cancel invitation

src/managers/InvitationManager/InvitationManagerStore.js
ipula: I will check this with @defstat

withanage:
@ipula  may be try to fix the tests also here.

public/globals.js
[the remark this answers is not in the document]
withanage: ok, then i would change

` 'invitation.cancelInvite.actionName':'Cancel Invite',` this to  ` 'invitation.cancelInvite.actionName':'Cancel Invitation'

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+		</div>
+	</div>
+	<PkpTable aria-label="Example for basic table">
+		<TableHeader>
+			<TableColumn>{{ t('invitation.tableHeader.name') }}</TableColumn>
+			<TableColumn>{{ t('about.contact.email') }}</TableColumn>
+			<TableColumn>{{ t('invitation.header') }}</TableColumn>
+			<TableColumn>{{ t('common.status') }}</TableColumn>
+			<TableColumn>{{ t('user.affiliation') }}</TableColumn>
+			<TableColumn></TableColumn>
```
jardakotesovec, line 22: For this one, lets use t('common.moreActions') with sr-only

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+	<div
+		v-if="!isInvitationLoading"
+		class="flex items-center pb-2 pt-2 text-3xl-normal"
+	>
+		<h4>
+			{{ t('invitation.header') }}({{ store.invitationsPagination.itemCount }})
+		</h4>
+		<div class="ml-auto">
+			<PkpButton @click="store.sendInvitation">
```
jardakotesovec, line 10: Very minor comment, but extra effort with naming makes the code easier to understand.

Here I would suggest createNewInvitation for example.

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+			<TableColumn></TableColumn>
+		</TableHeader>
+		<TableBody>
+			<TableRow v-for="(row, index) in store.invitations" :key="index">
+				<TableCell>
+					{{
+						row.userId
+							? row.existingUser.fullName
+							: store.getFullName(row.givenName, row.familyName)
+					}}
```
jardakotesovec, line 31: I think you are aware of this one - to move the combining the name to server side as there is more complex logic to that.

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+				</TableCell>
+				<TableCell>
+					{{ row.userId ? row.existingUser.email : row.email }}
+				</TableCell>
+				<TableCell>
+					<template v-for="(userGroups, i) in row.userGroupsToAdd" :key="i">
+						<span>
+							{{ localize(userGroups.userGroupName) }}
+						</span>
+						<br />
```
jardakotesovec, line 42: Even though I don't have good argument from top of my head for not using <br /> - have not see that for long time for controlling layouts :-).

Could we go with something bit more common. For example wrapping it in div with 'flex flex-col' classes I think should work.

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+					</template>
+				</TableCell>
+				<TableCell>
+					Invited
+					{{ new Date(row.createdAt).toLocaleDateString('en-US') }}
+				</TableCell>
```
jardakotesovec, line 48: Formatting dates is still on my list to improve for 3.5. For now please use the formatShortDate from '@/composables/useDate', which currently has very basic logic.. but its at least consistent and will improve that in future.

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+								icon: 'Cancel',
+								name: 'cancelInvite',
+								isWarnable: true,
+							},
+						]"
+						label="Invitation management options"
+						:display-as-ellipsis="true"
+						direction="left"
+						@action="(actionName) => store.handleReviewAction(row)"
+					/>
```
jardakotesovec, line 75: Please improve the naming here.  Something like handleInvitationAction for example would be more suitable.

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+								icon: 'Edit',
+							},
+							{
+								label: t('invitation.cancelInvite.actionName'),
+								icon: 'Cancel',
+								name: 'cancelInvite',
+								isWarnable: true,
+							},
+						]"
+						label="Invitation management options"
```
jardakotesovec, line 71: This needs to be localised

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+		const countPerPage = ref(2);
+		const currentPage = ref(1);
+		function setCurrentPage(_currentPage) {
+			currentPage.value = _currentPage;
+		}
+
+		const {apiUrl} = useApiUrl('invitations');
+		const getInvitationApiUrl = computed(() => {
+			return apiUrl.value;
+		});
```
jardakotesovec, line 46: I think you don't need create getInvitationApiUrl computed property.  

You can just
```
const {apiUrl: invitationApiUrl} = useApiUrl('invitations');
```

And than pass the invitationApiUrl to the useFetch

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+		/**
+		 * get all invited user groups
+		 * @param {Array} userGroups
+		 * @returns
+		 */
+		function getAllInvitedRoles(userGroups) {
+			let roles = '';
+			userGroups.forEach((element) => {
+				roles = roles + localize(element.userGroupName) + ', ';
+			});
```
jardakotesovec, line 86: Recently I noticed that we also have `common.commaListSeparator` in translations for `, ` :-). So trying to use that.

So instead of ', ' you can use t('common.commaListSeparator')

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+		function handleReviewAction(data) {
+			openDialog({
+				title: t('invitation.cancelInvite.title'),
+				message: t('invitation.cancelInvite.message', {
+					givenName: data.givenName,
+					familyName: data.familyName,
+					email: data.email,
+					roles: getAllInvitedRoles(data.userGroupsToAdd),
+					status:
+						'invited ' + new Date(data.createdAt).toLocaleDateString('en-US'),
```
jardakotesovec, line 100: Same thing, please use the formatShortDate

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+			});
+
+			return roles;
+		}
+
+		function handleReviewAction(data) {
+			openDialog({
+				title: t('invitation.cancelInvite.title'),
+				message: t('invitation.cancelInvite.message', {
+					givenName: data.givenName,
```
jardakotesovec, line 95: We are trying to keep the html mark up to absolute minimum in the translations. There is option in dialog to render the body via dedicated component if we need to display something more complex in the message than just text. Its exactly for these cases to avoid this inline html

Please checkout example in https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/story/components-dialog--with-body-component

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+		const {apiUrl} = useApiUrl('invitations');
+		const getInvitationApiUrl = computed(() => {
+			return apiUrl.value;
+		});
+		const {
+			items: invitations,
+			pagination: invitationsPagination,
+			isLoading: isInvitationLoading,
+			fetch: fetchInvitation,
+		} = useFetchPaginated(getInvitationApiUrl, {
```
jardakotesovec, line 52: Minor thing - but would be inclined to for example have fetchInvitations - so its clear its fetching the list, not individual invitation.

src/managers/InvitationManager/InvitationManagerStore.js
```diff
@@ -0,0 +1,146 @@
...
+							const {apiUrl: cancelApiUrl} = useUrl(
+								`invitations/${data.id}/cancel`,
+							);
+							const {fetch: cancelInvitation} = useFetch(cancelApiUrl.value, {
+								method: 'PUT',
+								body: {},
+							});
+
+							announce(t('common.loading'));
+							await cancelInvitation();
```
jardakotesovec, line 117: Looking good! I still don't have strong opinion on when/where to use the loading and would like to make it more automatic.  But it does not hurt to keep it here until I have better plan :-).

src/managers/InvitationManager/InvitationManager.vue
```diff
@@ -0,0 +1,107 @@
...
+		<h4>
+			{{ t('invitation.header') }}({{ store.invitationsPagination.itemCount }})
+		</h4>
+		<div class="ml-auto">
+			<PkpButton @click="store.sendInvitation">
+				{{ t('invitation.inviteToRole.btn') }}
+			</PkpButton>
+		</div>
+	</div>
+	<PkpTable aria-label="Example for basic table">
```
jardakotesovec, line 15: This aria-label is not ideal :-).

To make the headings/descriptions/buttons bit more consistent across these different tables, we introduced couple of slots where you can put thing without styling it should look fine - example https://github.com/pkp/ui-library/blob/main/src/managers/GalleyManager/GalleyManager.vue#L3

And that will also do correct accessibility automatically, which is connecting the title an description with the table.


It will look bit different compared to figma, but not worry about that. If we need this alternative styling we would add it to the table.

jardakotesovec (changes requested):
Hi Ipula,
I did review of InvitationManager. The others are still for me to pick up next days.

Thanks!

src/managers/InvitationManager/InvitationManager.mdx
```diff
@@ -0,0 +1,9 @@
+import {Primary, Controls, Stories, Meta, ArgTypes} from '@storybook/blocks';
+
+import * as InvitationManager from './InvitationManager.stories.js';
+
+<Meta of={InvitationManager} />
+
```
jardakotesovec, line 6: Bit more docs would be nice here.. :-)

Lets please rename this component to UserInvitationManager, as Alec was concerned that this suggests it would handle all invitations. But for example the invitations for reviewers will be likely handled elsewhere.

src/managers/InvitationManager/InvitationManager.vue
ipula: yes we have created a issue for that. I'm waiting for Dimithri's input here

src/managers/InvitationManager/InvitationManagerStore.js
ipula: I don't have an idea why I did this

src/managers/InvitationManager/InvitationManager.vue
asmecher: (link: https://github.com/craft-oa/gdpr-invitation/issues/17)

public/globals.js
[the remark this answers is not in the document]
ipula: image

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -14,9 +14,15 @@ export const useUserInvitationPageStore = defineComponentStore(
 		 * Invitation payload, initial value
 		 */
 		const invitationPayload = ref({...pageInitConfig.invitationPayload});
+		const updatedPayload = ref({});
```
jardakotesovec, line 17: Please add couple comments, so its clear whats the difference between invitationPayload and updatedPayload.

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -251,26 +291,65 @@ export const useUserInvitationPageStore = defineComponentStore(
...
+				],
+			});
+		}
+
 		const registeredActionsForSteps = {};
 		function registerActionForStepId(stepId, callback) {
 			registeredActionsForSteps[stepId] = callback;
 		}
 
+		function updateWithSelectedUserGroups(userGroups) {
```
jardakotesovec, line 339: I see that you do some updates directly in UserInvitationUserGroupsTable and some in store.

Original idea, was to keep the store somewhat general - so it should not know too much specifics about the payloads. So if possible to do the userGroup manipulations in UserInvitationUserGroupsTable?

src/managers/UserInvitationManager/UserInvitationManagerStore.js
```diff
@@ -0,0 +1,176 @@
...
+			window.location = sendInvitationPageUrl.value;
+		}
+
+		/**
+		 * get invitations with paginations
+		 */
+		const invitationCount = ref(0);
+
+		const countPerPage = ref(5);
+		const currentPage = ref(1);
```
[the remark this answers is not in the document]
withanage: @jardakotesovec , I discussed this with @Devika008  and she suggested 5 would be better.

src/pages/acceptInvitation/AcceptInvitationUserAccountDetails.vue
```diff
@@ -3,7 +3,7 @@
 		<div class="py-1">
 			<FormDisplayItemBasic
 				heading-element="h4"
-				:heading="t('user.emailAddress')"
+				:heading="t('user.email')"
 				:value="store.email"
 			></FormDisplayItemBasic>
 		</div>
```
jardakotesovec, line 9: Below I see multiple FieldText - would be possible to have these configured through the form as usual, rather than just individual Fields here?

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -111,21 +120,39 @@ import PkpButton from '@/components/Button/Button.vue';
 import FieldText from '@/components/Form/fields/FieldText.vue';
 import {useUserInvitationPageStore} from './UserInvitationPageStore';
 
+const props = defineProps({
```
jardakotesovec, line 123: Can't comment directly there because its not part of this PR.

But can Add Role button should not be part of the table.

You can use bottom-controls slot on the table to add it instead.

src/managers/UserInvitationManager/UserInvitationManagerCancelInvitationDialogBody.vue
```diff
@@ -0,0 +1,25 @@
...
+			<li>{{ t('userInvitation.roleTable.role') }} : {{ roles }}</li>
+			<li>{{ t('common.status') }} : {{ status }}</li>
+			<li>{{ t('user.affiliation') }} : {{ affiliation }}</li>
+		</ul>
+	</div>
+</template>
+
+<script setup>
+import {useTranslation} from '@/composables/useTranslation';
+
```
jardakotesovec, line 15: Please use useLocalize instead, useTranslation is deprecated. It works the same, so just fix the import.

src/managers/UserInvitationManager/UserInvitationManagerStore.js
```diff
@@ -0,0 +1,180 @@
...
+import {ref, watch} from 'vue';
+import UserInvitationManagerCancelInvitationDialogBody from './UserInvitationManagerCancelInvitationDialogBody.vue';
+
+export const useUserInvitationManagerStore = defineComponentStore(
+	'userInvitationsPage',
+	() => {
+		const {openDialog} = useModal();
+		const {localize} = useLocalize();
+		const {t} = useTranslation();
+		const {formatShortDate} = useDate();
```
jardakotesovec, line 20: useLocalize provides 't' as well..  I need to delete useTranslation after making sure its all moved to useLocalize. Thanks!

src/pages/userInvitation/UserInvitationSearchFormStep.vue
```diff
@@ -35,6 +35,7 @@ import FieldText from '@/components/Form/fields/FieldText.vue';
 import {useUserInvitationPageStore} from './UserInvitationPageStore';
 import {useFetch} from '@/composables/useFetch';
 import {useUrl} from '@/composables/useUrl';
+import {useTranslation} from '@/composables/useTranslation';
 
 defineProps({
 	validateFields: {
```
jardakotesovec, line 41: Does not seems to be used?

ipula: it used to the messages

src/pages/userInvitation/UserInvitationSearchFormStep.vue
```diff
@@ -78,21 +80,48 @@ async function searchUser() {
...
-			store.updatePayload('userId', user.id);
-			// we need to pass the givenName and familyName from API!!
-			store.updatePayload('givenName', user.fullName.split(' ')[0]);
-			store.updatePayload('familyName', user.fullName.split(' ')[1]);
-			store.updatePayload('orcid', user.orcid);
-			store.updatePayload('currentUserGroups', user.groups);
-			store.updatePayload('affiliation', user.affiliation);
+			let user = {};
+			if (userData.value.items.length > 1 && fields.value.username.length > 0) {
+				user = userData.value.items.find(
```
jardakotesovec, line 85: As mentioned on mattermost - all email/orcid and username are unique - so we have treat them the same when checking whether the result is unique.

src/pages/userInvitation/UserInvitationSearchFormStep.vue
```diff
@@ -78,21 +80,48 @@ async function searchUser() {
...
+					store.updatePayload('orcid', user.orcid);
+					store.updatePayload('currentUserGroups', user.groups);
+					store.updatePayload('affiliation', user.affiliation);
+					store.updatePayload('country', user.country);
+					store.userSearch.message = t('userInvitation.search.userFound');
+					store.userSearch.class = 'font-bold text-success';
+				}
+			} else {
+				user = userData.value.items[0];
+				store.updatePayload('inviteeEmail', user.email);
```
jardakotesovec, line 106: Please improve the structure - so the logic for updatePayload when user is found is not repeated.

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
```diff
@@ -10,8 +10,8 @@
 		<div class="p-1">
 			<FormDisplayItemBasic
 				heading-element="h4"
-				:heading="t('user.emailAddress')"
-				:value="store.invitationPayload.email"
+				:heading="t('user.email')"
+				:value="store.invitationPayload.inviteeEmail"
```
jardakotesovec, line 14: Don't have currently better idea other than using FormDisplay..  I think its actually not that bad :-).  Lets just fix the spacing between each of them. You can set probably something like this on the div: `flex flex-col gap-y-10`

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -14,9 +14,18 @@ export const useUserInvitationPageStore = defineComponentStore(
...
+		const updatedPayload = ref({});
+		const invitationType = ref(pageInitConfig.invitationType);
+		const invitationMode = ref(pageInitConfig.invitationMode);
 
-		function updatePayload(fieldName, value) {
+		function updatePayload(fieldName, value, initialValue = true) {
 			invitationPayload.value[fieldName] = value;
+			if (!initialValue) {
+				updatedPayload.value[fieldName] = value;
+			}
```
jardakotesovec, line 28: Can you help me understand why when its initial value it should not be included in updatedPayload.  This is probably correct..  but not intuitive.. so some additional comment would be useful.

ipula: if user did not update those values It will used in request payload. We only use user updated values on for the api request

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
```diff
@@ -85,17 +97,56 @@ const {
...
+				store.invitationPayload[field.name]
+					? store.invitationPayload[field.name]
+					: field.value,
+				store.invitationPayload[field.name] ? false : true,
+			);
+		} else {
+			if (store.invitationPayload[field.name] === null) {
+				store.updatePayload(field.name, field.value, true);
+			} else {
+				store.updatePayload(
```
jardakotesovec, line 121: Updating invitation payload with the value it already has? Does it do anything? Or maybe could be removed?

src/pages/userInvitation/UserInvitationEmailComposerStep.vue
```diff
@@ -27,8 +27,8 @@
 		:locale="props.email.locale"
 		:locales="props.email.locales"
 		:more-search-results-label="t('common.numberedMore')"
-		:recipient-options="props.email.recipientOptions"
```
jardakotesovec, line 30: Can't make comment above, as its not part of the PR. But we don't support attaching files in context sending invite, right? So the props like   attach-files-label, attachers, attachments, ...  can be removed.

ipula: I will remove those

src/pages/userInvitation/UserInvitationEmailComposerStep.vue
```diff
@@ -78,6 +78,21 @@ if (!store.invitationPayload.body) {
...
+	return [
+		{
+			value: store.invitationPayload.inviteeEmail,
+			label: {
+				[store.primaryLocale]: store.invitationPayload.inviteeEmail,
+			},
+		},
+	];
+});
+
```
jardakotesovec, line 95: I think we don't need filter by validateFields anymore?

ipula: I will check I used this most cases

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -47,26 +50,71 @@ export const useAcceptInvitationPageStore = defineComponentStore(
...
+					data.value.country,
+					userId.value ? true : false,
+				);
+				updateAcceptInvitationPayload(
+					'username',
+					null,
+					userId.value ? true : false,
+				);
+				updateAcceptInvitationPayload(
+					'password',
```
jardakotesovec, line 114: I don't think that password is being passed from api.

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -282,6 +304,34 @@ export const useAcceptInvitationPageStore = defineComponentStore(
...
+					}
+				});
+				payload = {
+					...updatedPayload.value,
+				};
+			} else {
+				currentStep.value.sections.forEach((element, index) => {
+					let sectionPayload = {};
+					element.props.validateFields.forEach((field) => {
+						if (Object.keys(updatedPayload.value).includes(field)) {
```
jardakotesovec, line 322: Do we still need to check the validateFields? If we are sending only the updated fields?

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -282,6 +304,34 @@ export const useAcceptInvitationPageStore = defineComponentStore(
 			);
 		});
 
+		const invitationRequestPayload = computed(() => {
+			let payload = {};
+			if (userId.value) {
+				Object.keys(updatedPayload.value).forEach((element) => {
+					if (updatedPayload.value[element] === null) {
+						delete updatedPayload.value[element];
+					}
```
jardakotesovec, line 313: Can you give me please example when this would happen that updatedPayload would be null?

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -292,13 +342,19 @@ export const useAcceptInvitationPageStore = defineComponentStore(
 			const {validationError, fetch} = useFetch(updateInvitationUrl, {
 				expectValidationError: true,
 				method: 'PUT',
-				body: updatedPayload.value,
+				body: {invitationData: invitationRequestPayload.value},
 			});
-			await fetch();
-			if (validationError.value) {
-				errors.value = validationError.value;
+			if (!acceptinvitationPayload.value.privacyStatement) {
```
jardakotesovec, line 347: please fix camelcase: acceptinvitationPayload -> acceptInvitationPayload

src/pages/acceptInvitation/AcceptInvitationPageStore.js
ipula: But in the mockups familyname country etc.. are required therefor we need to add those somehow.

src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -338,7 +396,8 @@ export const useAcceptInvitationPageStore = defineComponentStore(
 							{
 								label: t('acceptInvitation.modal.button'),
 								callback: (close) => {
-									close();
+									const {redirectToPage} = useUrl('submissions');
```
jardakotesovec, line 399: Just asked @defstat about what error scenario might happen on finalise - ideally it would be handled as well - and we would show error. Please follow-up with @defstat on some particular scenario that could be tested.

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -43,27 +31,89 @@
...
-					<template v-if="step.id === 'userDetails'">
-						<div v-for="section in step.sections" :key="section.id">
-							<FormDisplay v-bind="section.props.form"></FormDisplay>
+			</div>
+		</div>
+		<div v-else>
+			<div v-for="step in store.formSteps" :key="step.id">
+				<template v-if="step.id === 'userDetails'">
+					<div v-for="section in step.sections" :key="section.id">
+						<div
```
jardakotesovec, line 40: I think that you already iterate over the sections array in the AcceptInvitationPage? So maybe unnecessary to do it here again?

ipula: I use like `step.sections[0]`. Is it ok?

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
ipula: I updated this now. Please have look after the commit

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -43,27 +31,89 @@
...
+							>
+								<h3>{{ t('acceptInvitation.review.userDetails') }}</h3>
+								<PkpButton @click="store.openStep('userDetails')">
+									{{ t('common.edit') }}
+								</PkpButton>
+							</div>
+							<div class="p-4">
+								<div>
+									<h4>{{ locale.label }}</h4>
+								</div>
```
jardakotesovec, line 57: Locale and individual fields should not have same heading level. So lets move the fields to h5 for now.

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -43,27 +31,89 @@
...
+							:key="locale.key"
+							class="border-t border-light p-8"
+						>
+							<div
+								v-if="index === 0"
+								class="flex items-center justify-between p-4"
+							>
+								<h3>{{ t('acceptInvitation.review.userDetails') }}</h3>
+								<PkpButton @click="store.openStep('userDetails')">
+									{{ t('common.edit') }}
```
jardakotesovec, line 51: Please move the heading and edit button outside of the loop, as its before the individual locales start. It will be than easier to replace eventually the loop with locales with FormDisplay or something.

src/pages/acceptInvitation/AcceptInvitationReview.vue
```diff
@@ -43,27 +31,89 @@
...
+								</PkpButton>
+							</div>
+							<div class="p-4">
+								<div>
+									<h4>{{ locale.label }}</h4>
+								</div>
+
+								<FormDisplayItemBasic
+									v-if="store.userId === null"
```
jardakotesovec, line 60: I added multilingual example for FormDisplay to storybook on main branch..   So that now supports multilingual form, but its not flexible enough yet to display some of the custom parts here.

But you should be able to use directly the FieldTextDisplay and FieldSelectDisplay,

FieldTextDisplay now supports displayLocale prop.

ipula: I will use that

src/pages/acceptInvitation/AcceptInvitationUserAccountDetails.vue
```diff
@@ -55,19 +55,24 @@ import FormDisplayItemBasic from '@/components/FormDisplay/FormDisplayItemBasic.
...
 const options = [
 	{
 		value: true,
 		label:
 			t('acceptInvitation.privacyStatement.label') +
-			' <button class="text-blue-600 text-primary hover:underline"> ' +
+			' <a href=' +
+			pageUrl.value +
+			'><button class="text-blue-600 text-primary hover:underline"> ' +
 			t('acceptInvitation.privacyStatement.btn') +
```
jardakotesovec, line 74: a element should suffice, no need for additional button. If its link its better to just use the 'a' element if possible.

src/pages/acceptInvitation/AcceptInvitationPageStore.js
jardakotesovec: Right.. I still have bit unclear understanding how exactly the updatedPayload works. 

I thought that your intention was to update updatedPayload when given section is displayed.. to ensure that all metadata are included there? In other words that you rely on the fact that everything on the page which is there either from the start or updated from user ends up in updatedPayload - and thats what you send to server.  

Therefore I am bit confused what validateFields are addressing here?

src/pages/acceptInvitation/AcceptInvitationReview.vue
jardakotesovec: Is intention here just to find the form from the userDetails to reuse it?

If thats the case I would suggest just pass this form to both steps, so its available to both independently.

PR #384, opened 2024-07-16
pkp/pkp-lib#9468 exclude chapters without landing page and DOI from D…

bozana:
…OI list panel

s. https://github.com/pkp/pkp-lib/issues/9468

### DIFF
--- a/src/components/ListPanel/doi/DoiListItem.vue
+++ b/src/components/ListPanel/doi/DoiListItem.vue
@@ -69,7 +69,9 @@
 			<pkp-table :columns="doiListColumns" :rows="currentVersionDoiObjects">
 				<template slot-scope="{row}">
 					<table-cell :column="doiListColumns[0]" :row="row">
-						<label :for="row.uid">{{ row.displayType }}</label>
+						<label :for="row.uid" v-bind:class="{labelDisabled: row.disabled}">
+							{{ row.displayType }}
+						</label>
 					</table-cell>
 					<table-cell :column="doiListColumns[1]" :row="row">
 						<input
@@ -77,6 +79,7 @@
 							:id="row.uid"
 							type="text"
 							:readonly="!(isEditingDois && !isSaving)"
+							:disabled="isEditingDois && row.disabled"
 							v-model="
 								mutableDois.find((doi) => doi.uid === row.uid).identifier
 							"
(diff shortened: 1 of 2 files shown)


src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -500,6 +514,17 @@ export default {
 			);
 			return !hasAnyDois;
 		},
+		/**
+		 * Whether this object has chpaters that are disabled for DOI assignemnt
```
ewhanson, line 518: Typo in `chapters` and `assignment`

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -500,6 +514,17 @@ export default {
...
+		 *
+		 * @return {boolean}
+		 */
+		containsDisabledChapters() {
```
ewhanson, line 522: I can't think of a good way to do this without making things too complicated, so I'm curious about your thoughts on this: I think it would be preferable if all OMP specific code could be in `DoiListPanelOMP.vue` rather than being here. I don't know if there is a good way to do this and maybe that would just wildly complicate things to change it so maybe this is the best way. What do you think? Maybe keep it like this and if there start to be more application specific things that creep into the HTML itself, we can look into refactoring it?

bozana: I moved it now to DoiListPanelOMP...

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -77,13 +77,20 @@
...
+							:disabled="isEditingDois && row.disabled"
 							v-model="
 								mutableDois.find((doi) => doi.uid === row.uid).identifier
 							"
 						/>
 					</table-cell>
 				</template>
 			</pkp-table>
+			<span
+				class="doiListItem__disabledChaptersDescription"
```
ewhanson, line 89: I don't see anywhere this class is used at a quick glance, but if it is still needed, maybe remove reference to chapters in the class name?

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -230,13 +237,20 @@
...
+								:disabled="isEditingDois && row.disabled"
 								v-model="
 									mutableDois.find((doi) => doi.uid === row.uid).identifier
 								"
 							/>
 						</table-cell>
 					</template>
 				</pkp-table>
+				<span
+					class="doiListItem__disabledChaptersDescription"
```
ewhanson, line 249: Same about 'chatpers' in class name.

src/components/ListPanel/doi/DoiListItem.vue
bozana: Oh yes, I will remove it because we apparently do not need to style it...
Good catch! 
Thanks!

PR #389, opened 2024-07-26
pkp/pkp-lib#9744 Migrate older Tables to newer Table API

blesildaramirez:


src/components/Container/StatsPage.stories.js
```diff
@@ -115,28 +115,20 @@ const EditorialStatsPageWithDataAndTemplate = {
...
+						<TableHeader>
+							<TableColumn v-for="(column, columnIndex) in tableColumns" :key="column.name" :id="column.name">
+								{{column.label}}
+							</TableColumn>
+						</TableHeader>
+						<TableBody>
+							<TableRow v-for="(row, index) in tableRows" :key="index">
+								<TableCell>{{ row.name }}</TableCell>
+								<TableCell>{{ row.dateRange }}</TableCell>
+								<TableCell>{{ row.total }}</TableCell>
```
jardakotesovec, line 129: I see that in previous code there was tooltip included also?

src/components/Container/StatsPage.stories.js
```diff
@@ -115,28 +115,20 @@ const EditorialStatsPageWithDataAndTemplate = {
 				<div class="pkpStats__table" role="region" aria-live="polite">
 					<PkpTable
 						class="pkpTable--editorialStats"
-						labelled-by="editorialActivityTabelLabel"
-						:columns="tableColumns"
-						:rows="tableRows"
+						aria-label="Trends"
```
jardakotesovec, line 118: Also please check whether the column sizing works well.. As there seems to be some style overrides outside of the table. Again this is something we are trying to get rid off..   Instead of we are putting additional props and variants on components to be able handle different layout behaviour.

If you would conclude for example on this one that it requires some addition to the new Table, let me know. You can for example see that I already added couple of props to TableCell, like `noWrap` or `fullWidthTruncated` to handle some of the layout scenarios I had to deal with.
Screenshot 2024-08-07 at 13 55 08

src/components/TableNext/TableCell.vue
```diff
@@ -2,7 +2,7 @@
 	<component
 		:is="isRowHeader ? 'th' : 'td'"
 		:scope="isRowHeader ? 'row' : false"
-		class="border-b border-t border-light px-2 py-2 text-start first:border-s first:ps-3 last:border-e last:pe-3"
+		class="pkpTable__cell border-b border-t border-light px-2 py-2 text-start first:border-s first:ps-3 last:border-e last:pe-3"
 		:class="classes"
```
jardakotesovec, line 6: Adding class from legacy table cell? If we need improve styling in new one - we would do it with tailwindcss..

src/components/TableNext/TableRow.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-	<tr class="border-separate border border-light even:bg-tertiary">
+	<tr
+		class="pkpTable__row border-separate border border-light even:bg-tertiary"
+	>
```
jardakotesovec, line 4: Same issue here.. not sure why it was exactly added?

src/components/Container/StatsPage.stories.js
blesildaramirez: Update here, I just added a class="!max-w-[8rem] truncate" directly on my use of TableColumn. It's not common to have long column names in the tables I migrated so I think this would be enough.

src/pages/jobs/FailedJobsPage.vue
```diff
@@ -0,0 +1,135 @@
...
+				:key="column.name"
+				:id="column.name"
+			>
+				{{ column.label }}
+			</TableColumn>
+		</TableHeader>
+		<TableBody>
+			<TableRow v-for="row in rows" :key="row.key">
+				<table-cell>{{ row.id }}</table-cell>
+				<table-cell>{{ row.displayName }}</table-cell>
```
jardakotesovec, line 27: Its interesting that this has not been caught by eslint to be TableCell.

src/pages/jobs/JobsPage.vue
```diff
@@ -0,0 +1,61 @@
...
+		PkpTable,
+		TableHeader,
+		TableColumn,
+		TableBody,
+		TableRow,
+		TableCell,
+		Pagination,
+	},
+	extends: JobsPage,
+};
```
jardakotesovec, line 60: I suspect that extending from 'JobsPage' from containers might get confusing - what about rename it to JobsPageBase and move that to pages/jobs ?

jardakotesovec (changes requested):
Just couple small things.. its very close to being merged :-)

src/components/Container/StatsPage.stories.js
jardakotesovec: @blesildaramirez Can you check whether these additional classes are still necessary?

If so - I still would like to look into better options than such class overrides..

src/pages/jobs/JobsPage.vue
jardakotesovec: Seems still not moved to pages/jobs?

src/components/Container/StatsPage.stories.js
blesildaramirez: Removing the added class somehow worked and doesn't overlap (like from your screenshot above). So best to remove the class override. Thanks!

jardakotesovec (approved):

PR #391, opened 2024-08-12
pkp/pkp-lib#9890 Implement Collapsible Primary navigation

blesildaramirez:


src/components/Button/Button.vue
```diff
@@ -13,12 +13,16 @@
 		@blur="emit('blur')"
 	>
 		<Icon v-if="icon" class="h-5 w-5" :icon="icon" />
-		<slot />
+		<span v-if="sizeVariant === 'iconOnly' && icon" class="sr-only">
+			<slot />
```
jardakotesovec, line 17: You mentioned in notes - that you think it will be used in future? Have you seen any example of that? 

Its not trivial addition, so just would like to make sure that its added only if we are confident that it will be used. If anything we are looking for removing things, to keep things simple  :-).

src/components/SideMenu/SideMenu.vue
```diff
@@ -0,0 +1,202 @@
...
+		:model="items"
+		:pt="navigationStyling"
+		class="w-max min-w-60 overflow-y-auto border-e border-s border-light bg-secondary"
+	>
+		<template #item="{item, active, hasSubmenu}">
+			<a
+				:class="getButtonStyles(item)"
+				:href="item.link"
+				tabindex="-1"
+				@click="handleClick(item)"
```
jardakotesovec, line 13: Just wondering why tabindex -1 would be used here? In their examples they also use 'a' without adjusting tabindex.

I had difficulties to using key up/down to navigate through the menu. But maybe it has something to do with storybook, not sure. But one thing we need to make sure that intended accessibility of PanelMenu works well.

src/components/Button/Button.vue
blesildaramirez: Hi Jarda, for now there are no usage of this on the button component. So I removed it. We can always come back and re-implement it if it becomes necessary in the future.

src/components/Button/Button.vue
```diff
@@ -103,4 +112,23 @@ const emit = defineEmits([
...
+});
 </script>
+
+<style lang="less" scoped>
+@import '../../styles/_import';
+
+a.text-on-dark:hover,
+a.text-on-dark:focus,
+a.text-on-dark:active {
+	color: rgb(255 255 255 / var(--tw-text-opacity));
```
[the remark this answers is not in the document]
blesildaramirez: I moved this css rules on the SideMenu component. We need this because when there's a conflicting old css rule when this component is used in OJS, OMP and OPS.
image

src/components/SideMenu/SideMenu.vue
blesildaramirez: This is one of the accessibility problems I ran into when implementing this component. Will look into this further.

src/components/SideNav/SideNav.vue
```diff
@@ -0,0 +1,68 @@
...
+	return result;
+}
+</script>
+
+<style lang="less" scoped>
+@import '../../styles/_import';
+
+.nav-section {
+	height: calc(100vh - 3rem);
+}
```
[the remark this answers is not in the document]
blesildaramirez: This works, thanks!

src/components/SideMenu/SideMenu.vue
```diff
@@ -0,0 +1,212 @@
...
+
+<style lang="less" scoped>
+@import '../../styles/_import';
+
+a.text-on-dark:hover,
+a.text-on-dark:focus,
+a.text-on-dark:active {
+	color: rgb(255 255 255 / var(--tw-text-opacity));
+}
+</style>
```
jardakotesovec, line 212: Would be nice if we can handle this in tailwindcss..  let me know if you think its not practical to do so..

src/components/SideNav/SideNav.vue
```diff
@@ -0,0 +1,60 @@
...
+	},
+	/** Aria label to be set for the nav element */
+	ariaLabel: {
+		type: String,
+		default: 'Site Navigation',
+	},
+});
+
+const items = reactive(convertLinksToArray(props.links));
+
```
jardakotesovec, line 39: this might be more suitable for computed property? So it would get updated if the props.links gets updated?

src/components/Button/Button.vue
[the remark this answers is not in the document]
jardakotesovec: I see ok! Please add it as comment as reminder why its done like that.

src/components/SideMenu/SideMenu.vue
jardakotesovec: Oh nevermind.. Just read the comment above

PR #393, opened 2024-08-19
pkp/pkp-lib#9890 Add useSideMenu composable

blesildaramirez:


src/components/SideMenu/SideMenu.vue
```diff
@@ -61,38 +60,58 @@ const props = defineProps({
...
-// Maps the key and level attributes which are necessary to render the nested menu
-function mapItems(_items, level = 1, _key) {
+const localExpandedKeys = ref({...props.expandedKeys});
+watch(
+	() => props.expandedKeys,
+	(newExpandedKeys) => {
+		localExpandedKeys.value = {...newExpandedKeys};
+	},
+);
+
```
jardakotesovec, line 102: When managing state in components, I think its always good to answer clearly - who is owning the state. Sometime its parent sometime its the children. But it gets difficult if its something in between.

In this case SideMenu is owning the state, but you can impact it via props. When props gets changes it merges changes in props with original state and you get something in between.

What I had in mind is that parent would own the state using the useSideMenu composable. That would be source of the truth - and SideMenu component would just strictly display the state its receiving via props. Without handling its own state that would be synchronised with props.

That also allow more powerful api that I was asking for - like expandAll.

PR #398, opened 2024-08-26
pkp/pkp-lib#10263 Relax editing metadata on published/posted materials

Hafsa-Naeem:
fixes [#10263 ](https://github.com/pkp/pkp-lib/issues/10263)

### DIFF
--- a/src/components/Container/WorkflowPage.stories.js
+++ b/src/components/Container/WorkflowPage.stories.js
@@ -220,12 +220,6 @@ const WorkflowPageWithDataAndTemplate = {
 							</PkpButton>
 						</template>
 					</PkpHeader>
-					<div
-						v-if="workingPublication.status === getConstant('STATUS_PUBLISHED')"
-						class="pkpPublication__versionPublished"
-					>
-						This version has been published and can not be edited.
-					</div>
 					<tabs
 						class="pkpPublication__tabs"
 						:is-side-tabs="true"
--- a/src/components/Container/WorkflowPage.vue
+++ b/src/components/Container/WorkflowPage.vue
@@ -487,11 +487,9 @@ export default {
 					return field;
 				});
 
-				// Add/remove save button depending on publication status or user permissions
-				form.canSubmit =
-					this.canEditPublication &&
-					publication.status !== pkp.const.STATUS_PUBLISHED;
-
+				// Add/remove save button depending on user permissions
+				form.canSubmit = this.canEditPublication;
+				
 				// Pass the publication status to the issue selection field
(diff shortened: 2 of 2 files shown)


asmecher (changes requested):
The workflow still includes a Vue file called `WorkflowPublicationEditDisabled.vue`, but with these changes, it's used to present a warning, rather than a prohibition against editing. The filename should be adjusted -- but watch out in case this is also used in the author interface; I didn't check. If it is used there, then "disabled" is still appropriate; it might be necessary to either make a copy for the editor's interface under a different name, or generalize it to either warn or prohibit changes.

src/managers/GalleyManager/useGalleyManagerConfig.js
```diff
@@ -84,15 +84,6 @@ export function useGalleyManagerConfig() {
 	function getManagerConfig({submission, publication}) {
 		const permittedActions = GalleyManagerConfiguration.actions
 			.filter((action) => {
-				if (
-					publication.value.status === pkp.const.STATUS_PUBLISHED &&
-					GalleyManagerConfiguration.actionsRequiresUnpublishedState.includes(
```
asmecher, line 89: This removes the only use of `actionsRequiresUnpublishedState` higher up in the file. If we're removing this reference, then that list should be removed too.

src/pages/workflow/components/publication/WorkflowPublicationEditDisabled.vue
```diff
@@ -2,7 +2,7 @@
 	<div
 		class="bg -mx-5 -mt-5 bg-attention p-5 text-center text-base-normal text-on-dark"
 	>
-		{{ t('publication.editDisabled') }}
+		{{ t('publication.editorEditWarning') }}
```
asmecher, line 5: I think this change is accidental -- it should still be `editDisabled` for this file.

src/pages/workflow/composables/useWorkflowConfig/workflowConfigEditorialOJS.js
```diff
@@ -930,7 +930,7 @@ export const PublicationConfig = {
 						formName: 'identifier',
 						submission,
 						publication: selectedPublication,
-						canEdit: permissions.canEditPublication,
+						canEdit: 'true',
```
asmecher, line 933: This is getting hard-wired `true`, but there are lots of other references to `canEdit` in this file that still use `permissions.canEditPublication` (and that property still appears to be getting supplied). Is this change necessary?

asmecher (changes requested):
A few changes suggested here!

PR #404, opened 2024-08-29
pkp/pkp-lib#10350 fix the check of the COUNTER R5 form field customer_id

bozana:
s. https://github.com/pkp/pkp-lib/issues/10350


jardakotesovec (approved):
Yes, thats good. Confirmed on mattermost that customer_id would be always integer in our system. With that assumption I think this is accurate.

PR #408, opened 2024-09-09
pkp/pkp-lib#9890 Update the design for the focused item state in SideMenu

blesildaramirez:


src/components/SideNav/SideNav.vue
```diff
@@ -94,9 +99,12 @@ function getExpandedKeys(items) {
...
-	getExpandedKeys(items),
-);
+const {sideMenuProps} = useSideMenu(items, {
+	activeItemKey: currentActiveKey,
+	expandedKeys: getExpandedKeys(items),
+});
+
+watchEffect(() => {
+	Object.assign(items, convertLinksToArray(linksRef.value));
+});
```
jardakotesovec, line 109: This is because of some runtime menu changes, right? Can you just explain bit more how it works?

jardakotesovec (approved):
Looking great, only one question to better understand how the dynamic changes of the menu works.

src/components/SideNav/SideNav.vue
blesildaramirez: Hi Jarda, sorry I wasn't able to provide explanation prior merging, I didn't notice you had a question here.
The props coming from php are not propagated down to the useSideMenu API so this change was necessary. An example of this runtime change is when enabling the Announcements feature where we don't reload the page and just add the link to the sidebar.
As for the `watchEffect`, when `props.links` get updated, then it was necessary to update the `items` array so the `useSideMenu` can propagate the changes down to the PanelMenu component. If you have other suggestion to make this work better, then please let me know and I can get the changes on a separate PR. Thanks!

PR #413, opened 2024-09-19
pkp/pkp-lib#10447 Create new ButtonIcon component

blesildaramirez:


blesildaramirez:
Code review notes:
1. I included new icons in the PR, although should be address on a separate one but I did not include all new icons from Figma. I'll do the rest of additions soon for https://github.com/pkp/pkp-lib/issues/9626
2. Failing test for `DropdownActions` component, this is because I removed the `<div class="sr-only"></div>` and used `aria-label` for the `<button>` instead.
3. Removed fixed width and height on Expand and History icons.

PR #414, opened 2024-09-24
pkp/pkp-lib#9626 Migrate icons to svg

blesildaramirez:


jardakotesovec (approved):
Looking great! Thank you for going through all these cases!

PR #415, opened 2024-09-25
pkp/pkp-lib#10444 Improve dialog component design

blesildaramirez:


src/components/Container/ManageEmailsPage.vue
```diff
@@ -325,7 +328,7 @@ export default {
 						title: mailable ? mailable.name : '',
 						mailable: this.currentMailable,
 						onOpenTemplate: this.openTemplate,
-						onConfirmResetTemplate: this.confirmRemoveTemplate,
+						onConfirmResetTemplate: this.confirmResetTemplate,
```
blesildaramirez, line 331: Note: This is supposed to open this modal (`confirmResetTemplate`) to confirm resetting an email template. Including this fix on this PR.

src/components/Modal/Dialog.vue
```diff
@@ -116,16 +124,61 @@ const props = defineProps({
...
+	/** Defines if clicking outside the modal should close it */
+	isDismissible: {
+		type: Boolean,
+		default: () => true,
+	},
+});
+
+const styles = computed(() => ({
+	'modal__panel modal__panel--dialog relative mx-3 w-10/12 max-w-3xl transform overflow-hidden rounded bg-secondary text-start shadow transition-all sm:my-8': true,
+	'border-none': props.modalStyle === 'basic',
```
jardakotesovec, line 143: modal__panel and modal__panel--dialog are there only for e2e tests? Or also contain some styling? If there would be opportunity to removes these I would not mind :-)

jardakotesovec: Btw Dialog can be referenced as [data-cy=dialog] in e2e now.

src/components/Modal/Dialog.vue
```diff
@@ -116,16 +124,61 @@ const props = defineProps({
...
+
 const emit = defineEmits(['close']);
 
 const isLoading = ref(false);
 
-function onClose() {
+function onClose(evtType) {
+	// Prevent closing the modal on outside click or "Esc" key if isDismissible is false
+	if (!props.isDismissible && evtType === 'default') {
+		return;
```
jardakotesovec, line 175: Could you please explain what that evtType === 'default' means?

jardakotesovec (approved):
Just two minor requests/questions. But looking great! Lets merge it soon :-).

src/components/Modal/Dialog.vue
blesildaramirez: Hi @jardakotesovec , we call this function when we want to close the modal. At this point, I added a param `evtType` with value "default" in this scenario:
```
<HLDialog class="modal" @close="onClose('default')">
```
This `@close` event is triggered when the user clicks outside of the modal, or pressing "Esc" key. And if the modal `isDismissble` is true, then we should skip closing the modal, thus only allowing closing the modal via clicking "Close" or "Cancel" buttons.

PR #417, opened 2024-09-27
pkp/pkp-lib#7135 Multiple author affiliations

GaziYucel:
Related issues:
- https://github.com/pkp/pkp-lib/issues/7135

Pull requests:
- https://github.com/pkp/pkp-lib/pull/10460
- https://github.com/pkp/ui-library/pull/417
- https://github.com/pkp/ojs/pull/4469
- https://github.com/pkp/omp/pull/1819
- https://github.com/pkp/googleScholar/pull/23
- https://github.com/pkp/jatsTemplate/pull/68
- https://github.com/asmecher/grobidMetadataExtractor/pull/1
- https://github.com/pkp/quickSubmit/pull/102


src/components/Form/fields/FieldAffiliations.stories.js
```diff
@@ -0,0 +1,438 @@
...
+			Icon,
+			t
+		},
+		setup() {
+			// Storybook
+			const isStoryBook = true;
+			const currentValue = args.items;
+			const organizations = ref(args.apiResponse['items']);
+			const currentLocale = args.currentLocale;
+			const supportedLocales = args.supportedLocales;
```
jardakotesovec, line 38: Stories should not contain the logic from the component itself. Just imported and use it and pass props to it. 

Could can check out other Form fields for example.

GaziYucel:
I am waiting for a variant of the FieldBaseSuggest.vue component. After this, this PR can be wrapped up.

GaziYucel:
Hi @jardakotesovec,

I have implemented everything we discussed and all seems to work. 

There are some notes:
- you mentioned that Blessie was working on a variant of FieldBaseSuggest.vue which I can use;
- I have used <input type="text" for input fields. It would be better to have a very simple Vue component to use. This would have the following advantages:
  - less clutter and cleaner code
  - better accessibility
- error handling
  -  this is not implemented a la OJS
  - the check I do: there should be a ROR or at least one name and the name should be the primary locale
  - if this is not true, I disable the 'Add' button
  - saving the form is still enabled
- UI design
  - @Devika008 used a pie to show how many translations there are. I couldn't do this, because there is no component for this. I implemented "MultilingualProgress" instead
  - Devika placed the text/link "# of # translations completed" below the input fields. I moved this to above, because this makes the interface less jumpy

GaziYucel:
Hi @jardakotesovec,

I have implemented the new FieldRorAutosuggest (FieldAffiliationsRorAutoSuggest) component of @blesildaramirez. 

I think the PR is ready for review now. Thanks. 

@bozana: fyi

GaziYucel:
I am copying this from Mattermost to here, so it's on the list to check. 

As mentioned before, I think AbortController in src/composables/useFetch.js is not working as intended. Removing or adding async / await does not help. 

You can recreate this with the following two files, which are also in this PR. 

* src/components/Form/fields/FieldAffiliationsTest.stories.js
* src/components/Form/fields/FieldAffiliationsTest.vue

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+								<span class="text-lg-semibold">
+									{{ affiliation.name[primaryLocale] }}
+								</span>
+								<a
+									v-if="affiliation.ror"
+									:href="affiliation.ror"
+									target="_blank"
+								>
+									<Icon :icon="'ROR'" :class="'ml-1 h-6 w-6'" :inline="true" />
+								</a>
```
jardakotesovec, line 42: Minor suggestion, just fyi, when passing some text constant to the props, you can just skip the colon and do `icon="ROR"` class="ml-1 h-6 w-6"

Instead of ml-1 probably ms-1 is better as that should work for RTL layout as well.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+									<div
+										v-for="[newAffiliationNameLocale] in Object.entries(
+											newAffiliation.name,
+										)"
+										:key="newAffiliationNameLocale"
+									>
+										<div
+											v-if="supportedLocales.includes(newAffiliationNameLocale)"
+										>
+											<div>
```
jardakotesovec, line 180: I think here the FieldText also should work well here. As there is lots being replicated from that component anyway.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+												/>
+												<FieldError
+													v-if="
+														newAffiliationNameLocale === primaryLocale &&
+														!newAffiliation.name[
+															newAffiliationNameLocale
+														].trim()
+													"
+													:messages="[t('validator.required', {})]"
+												/>
```
jardakotesovec, line 217: Here to avoid custom error logic - I would suggest just to indicate that its required and if someone deletes it maybe just disabling the add button would be obvious enough. But thats just my opinion, its probably good either way.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+											</div>
+										</div>
+									</div>
+								</div>
+							</div>
+						</TableCell>
+						<TableCell>
+							<div v-if="showNewAffiliationForm">
+								<button
+									class="pkpButton border-transparent py-2 text-lg-semibold"
```
jardakotesovec, line 227: Any reason not to use our Button.vue component? Similarly like with the table wi intend to use our Button to ensure consistent rendering, rather than native component.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+		return [];
+	}
+	return props.allErrors[props.name];
+});
+
+watch(
+	currentValue,
+	() => {
+		makeCurrentValueCompatible();
+	},
```
jardakotesovec, line 348: What scenario this should address? Its bit intricate to reason about this, as its changing currentValue when the currentValue got changed. I think only reason why its not infinite loop is because the makeCurrentValueCompatible is not updating the 'value', but inline fields.

Would like to have cleaner solution.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+		message: t('user.affiliations.deleteModal.message', {
+			affiliation: currentValue.value[index]['name'][primaryLocale],
+		}),
+		actions: [
+			{
+				label: t('common.yes', {}),
+				isWarnable: true,
+				callback: async (close) => {
+					currentValue.value.splice(index, 1);
+					close();
```
jardakotesovec, line 476: I suspect that these in-object edits don't trigger the setter, which emits the change to the form. 

Better practice (chatgpt can provide more detailed  explanation) can be
```
  currentValue.value = currentValue.value.filter((_, i) => i !== index);

``

File: src/components/Form/fields/FieldAffiliations.vue [hunk 8]
```diff
@@ -0,0 +1,609 @@
...
+				...affiliation.name,
+				[locale]: value,
+			},
+		};
+	});
+}
+
+// helper methods
+
+function makeCurrentValueCompatible() {
```
jardakotesovec, line 508: Please add comment to explain what scenario is this function addressing.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+
+<style lang="less">
+@import '../../../styles/_import';
+
+.pkpFormField--affiliations {
+	position: relative;
+}
+
+.pkpFormField--affiliations__control {
+	position: relative;
```
jardakotesovec, line 590: I see that for most cases the tailwindcss is used, which is now preferred method. Can we also move these to the tailwindcss?

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,609 @@
...
+});
+const authorId = props.authorId;
+const primaryLocale = props.primaryLocale;
+const locales = props.locales;
+const currentValue = computed({
+	get: () => props.value,
+	set: (newVal) => emit('change', props.name, 'value', newVal),
+});
+const supportedLocales = props.locales.map((language) => language.key);
+const defaultLocale = 'en';
```
jardakotesovec, line 317: How this defaultLocale exactly works? I see its used
```
							<span class="text-lg-semibold">
								{{
									t('user.affiliations.searchPhraseLabel', {
										language: getLocaleDisplayName(defaultLocale),
									})
								}}
							</span>

```
Which suggest that if english is not primary language that it would not work? Can you check?

jardakotesovec (changes requested):

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: icon="ROR" doesn't work, only :icon="ROR" works. You can quickly see this in storybook.

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: The problem I had with FieldText is that the value is not binded two ways. I see that you replaced my "input" with FieldText in the affiliations loop. Look closely there, it is not working. The binding is only one way, thus showing but the object is not updated. That's what I meant that I couldn't make it work without changes in FieldText component.

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: The main goal here is to get all locales correct when the affiliations comes from the database. So I think it would suffice to run this only once at onMounted. What do you think, would this then be okay for you?

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: The label for the search phrase should state that the search phrase should be input in English. That's why I have 'en' as the locale there. This is how @Devika008 wanted this.

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: Found why, notice the difference: `icon="ROR"` <> `:icon="'ROR'"`
Without the colon, the value should be ROR, not 'ROR'. Sorry.

src/components/Form/fields/FieldAffiliations.vue
bozana: I think this should be the submission primary locale -- it does not need to be English -- the journal could not have English at all.

bozana: And by the way: it seems the contributor form does not handle/display the submission primary locale correctly, s https://github.com/pkp/pkp-lib/issues/10879.

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: I removed this method completely. At onMounted(), the names are only sorted now with the primary first. 
Apparently, this is already done in the backend in the class Schema.

src/components/Form/fields/FieldAffiliations.vue
GaziYucel: This method is removed.

src/components/ListPanel/contributors/ContributorsListPanel.vue
```diff
@@ -294,6 +294,13 @@ export default {
 			let activeForm = cloneDeep(this.form);
 			activeForm.action = this.contributorsApiUrl;
 			activeForm.method = 'POST';
+			activeForm.fields = activeForm.fields.map((field) => {
+				if (field.name === 'affiliations') {
+					field.primaryLocale = activeForm.primaryLocale;
+					field.locales = activeForm.supportedFormLocales;
+				}
```
jardakotesovec, line 301: Do we need to copy this from form to the field? By looking to FieldBase I see that it should be getting these two information as props - primaryLocale and locales.

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,549 @@
...
+										</div>
+									</div>
+								</div>
+							</div>
+						</TableCell>
+						<TableCell class="text-right">
+							<DropdownActions
+								v-if="!(affiliationIndex === indexEditMode)"
+								v-bind="rowActionsArgs(affiliationIndex)"
+								@action="rowActionsHandler"
```
jardakotesovec, line 122: Instead of changing DropdownActions with additional property of `id`. Could you please just update it here:

(actionName) => rowActionsHandler(actionName, affiliationIndex)

src/components/Form/fields/FieldAffiliations.vue
```diff
@@ -0,0 +1,549 @@
...
+		actions: actions,
+		label: t('user.affiliations.translationActionsAriaLabel', {}),
+		ariaLabel: t('user.affiliations.translationActionsAriaLabel', {}),
+		direction: 'left',
+		buttonVariant: 'ellipsis',
+	};
+};
+
+function rowActionsHandler(param) {
+	if (typeof param === 'object' && param.length === 2) {
```
jardakotesovec, line 416: Could we just be explicit abyt the arguments here, like rowActionhandler(actionName, affiliationIndex)?

jardakotesovec (changes requested):
Thank you for the updates. Could you please check couple more? Should be easy updates.

src/components/ListPanel/contributors/ContributorsListPanel.vue
GaziYucel: How would I get these from FieldBase in my component FieldAffiliations? Can you help me with this?

jardakotesovec: This means that its being passed to all fields. Including your one. So if you just add it to the FieldAffiliation props - it should become available to you.

jardakotesovec (approved):
At some point we should do some screen reader accessibility testing. But don't have bandwidth for that at this point. 

I think its in good enough shape to be merged. Thank you for the hard work on this!

PR #426, opened 2024-10-14
pkp/pkp-lib#4787 Reviewer suggestions

touhidurabir:
for pkp/pkp-lib#4787

### DIFF
--- a/src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
+++ b/src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
@@ -101,6 +101,14 @@ export default {
 		 * @param
 		 */
 		select() {
+			if (this.item.existingUserId && this.item.existingReviewerRole) {
+				pkp.eventBus.$emit('selected:reviewer', {
+					id: this.item.existingUserId,
+					fullName: this.localize(this.item.fullName),
+				});
+				return;
+			}
+
 			const {t} = useLocalize();
 
 			const {openLegacyModal} = useLegacyGridUrl({
@@ -110,7 +118,11 @@ export default {
 					submissionId: this.submissionId,
 					stageId: this.stageId,
 					reviewRoundId: this.reviewRoundId,
-					selectionType: pkp.const.REVIEWER_SELECT_CREATE,
+					selectionType: this.item.existingUserId
+						? this.item.existingReviewerRole
+							? pkp.const.REVIEWER_SELECT_ADVANCED_SEARCH
+							: pkp.const.REVIEWER_SELECT_ENROLL_EXISTING
+						: pkp.const.REVIEWER_SELECT_CREATE,
 					reviewerSuggestionId: this.item.id,
 				},
 			});
--- a/src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
+++ b/src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
@@ -3,6 +3,7 @@
 		<div class="listPanel__itemSummary">
 			<div class="listPanel__itemIdentity">
 				<div class="listPanel__itemTitle">
+					<!-- TODO: check why localize(fullName) causing error -->
 					{{ fullName }}
 					<Badge class="listPanel__item--reviewer__active">
 						{{ affiliation }}
@@ -11,6 +12,7 @@
 
 				<div class="listPanel__itemSubtitle">
 					<div class="listPanel__item--reviewer__affiliation">
+						<!-- TODO: how to show without v-html -->
 						{{ suggestionReason }}
 					</div>
 				</div>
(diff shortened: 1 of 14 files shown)


src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
```diff
@@ -0,0 +1,125 @@
...
+					<!-- TODO: check why localize(fullName) causing error -->
+					{{ fullName }}
+				</div>
+
+				<div class="listPanel__itemSubtitle">
+					<div class="listPanel__item--reviewer__affiliation">
+						{{ affiliation }}
+					</div>
+					<!-- TODO: check alternative of v-html as v-strip-unsafe-html not working -->
+					<div v-html="suggestionReason"></div>
```
touhidurabir, line 15: Need better and secure replacement of `v-html` .

src/managers/ReviewerSuggestionManager/ReviewerSuggestionManager.vue
```diff
@@ -0,0 +1,71 @@
...
+							<div class="text-base-bold">
+								{{ reviewerSuggestion.fullName }}
+							</div>
+							<div class="text-sm-normal text-secondary">
+								{{ reviewerSuggestion.affiliation }}
+							</div>
+							<!-- TODO: check alternative of v-html as v-strip-unsafe-html not working -->
+							<div
+								class="text-sm-normal text-secondary"
+								v-html="reviewerSuggestion.suggestionReason"
```
touhidurabir, line 32: Need better and secure replacement of `v-html` .

defstat (changes requested):
Thanks @touhidurabir! Not much here.

Questions:

In the "REVIEWER SUGGESTIONS" there is a drop down control that has a "More Actions" indicator. Should that be a three-dot button doing the same thing? 

Image

Also I am wondering whether the newly upgraded TinyMCE affects in any way the forms that are being used in this feature, or rather whether we need to consider changes in this issues PRs regarding the upgraded TinyMCE component.

touhidurabir:
> Thanks @touhidurabir! Not much here.
> 
> Questions:
> 
> In the "REVIEWER SUGGESTIONS" there is a drop down control that has a "More Actions" indicator. Should that be a three-dot button doing the same thing?
> 
> Image
> 
> Also I am wondering whether the newly upgraded TinyMCE affects in any way the forms that are being used in this feature, or rather whether we need to consider changes in this issues PRs regarding the upgraded TinyMCE component.

This has broken due to some new changes in the ui lib , otherwise this is how it should be and was previously
CleanShot 2025-02-04 at 15 29 53@2x

src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
touhidurabir: This is related to https://github.com/pkp/pkp-lib/issues/9421 , where we have introduced a new directive `v-strip-unsafe-html` for 3.3/3.4 but has not yet pushed it for 3.5 . So this is just a placeholder to update it once this change or something similar get pushed for 3.5 .

src/managers/ReviewerSuggestionManager/ReviewerSuggestionManager.vue
touhidurabir: same as https://github.com/pkp/ui-library/pull/426#discussion_r1941002067

asmecher:
@blesildaramirez, it looks like some TinyMCE changes might be needed here -- can you spot anything quickly that needs adaptation for the new version?

touhidurabir:
> @blesildaramirez, it looks like some TinyMCE changes might be needed here -- can you spot anything quickly that needs adaptation for the new version?

This does not seems like anything related to TinyMce but some changes to the DropdownActions component and the fix is very simple .

asmecher:
> This does not seems like anything related to TinyMce but some changes to the DropdownActions component and the fix is very simple .

Thanks, @touhidurabir -- do you mean a simple fix that you can tackle, or that you'll need someone to take care of in `ui-library`?

touhidurabir:
> > This does not seems like anything related to TinyMce but some changes to the DropdownActions component and the fix is very simple .
> 
> Thanks, @touhidurabir -- do you mean a simple fix that you can tackle, or that you'll need someone to take care of in `ui-library`?

I have already fixed it . Working on the PR update based on code review .

src/components/ListPanel/users/SelectReviewerSuggestionListItem.vue
```diff
@@ -0,0 +1,155 @@
...
```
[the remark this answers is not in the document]
touhidurabir: Yes, we should but I think it should be the part of different task to have a general component to handle such cases instead only for this issue .

defstat (changes requested):
Thanks @touhidurabir! Can you check this scenario?

When in Review Round workflow, click on "Add Reviewer". In the modal that appears, click "Select Reviewer" from the "Select a Reviewer from Reviewer Suggestions" grid. In the "Add Reviewer" modal, try to click on "Files To Be Reviewed". The button seems not to work as it supposed to though.

touhidurabir:
> Thanks @touhidurabir! Can you check this scenario?
> 
> When in Review Round workflow, click on "Add Reviewer". In the modal that appears, click "Select Reviewer" from the "Select a Reviewer from Reviewer Suggestions" grid. In the "Add Reviewer" modal, try to click on "Files To Be Reviewed". The button seems not to work as it supposed to though.

great spot of this case as only happening from `Add Reviewer` modal of all reviewers and also again only for suggestion or existing user with no reviewer role . Should be fixed with last commit .

PR #429, opened 2024-10-18
pkp/pkp-lib#10529 Fix: orderable FieldOptions doesn't show correct order

NateWr:


blesildaramirez (approved):
Works as intended, thanks!

PR #433, opened 2024-10-24
pkp/pkp-lib#9884 Second attempt to override default behaviour for one…

jardakotesovec:
… input

### DIFF
--- a/src/components/Form/Form.vue
+++ b/src/components/Form/Form.vue
@@ -3,8 +3,9 @@
 		class="pkpForm -pkpClearfix"
 		:method="method"
 		:action="action"
-		@keydown.enter.prevent=""
+		@submit.prevent="() => {}"
 	>
+		<input type="submit" value="i9884" style="display: none" />
 		<input
 			v-for="(value, name) in hiddenFields"
 			:key="name"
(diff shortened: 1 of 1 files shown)


src/components/Form/Form.vue
```diff
@@ -3,8 +3,9 @@
 		class="pkpForm -pkpClearfix"
 		:method="method"
 		:action="action"
-		@keydown.enter.prevent=""
+		@submit.prevent="() => {}"
```
jonasraoni, line 6: AFAICR you can leave it empty, but doesn't matter.

src/components/Form/Form.vue
```diff
@@ -3,8 +3,9 @@
...
 	>
+		<input type="submit" value="i9884" style="display: none" />
```
jonasraoni, line 8: Perhaps this was forgotten.

jonasraoni (approved):

src/components/Form/Form.vue
jardakotesovec: Nope... its intentional.. its explained in the issue.  This is important for scenario, when there is one input, but some other input. Which might be "help tooltip" or remove button for individual pill in autocomplete.  Because than browser automatically clicks on first button regardless what it is..  So it for example starts removing pills on enter.    

So this is intentionally to be first button in form, which does nothing and which gets automatically clicked on by browser when there is just one input.

blesildaramirez (approved):

PR #437, opened 2024-10-28
pkp/pkp-lib#9658 user access table and table actions

ipula:


jardakotesovec:
@withanage @ipula Whats missing is to determine which actions would be available to current user. Here is the logic from original table https://github.com/pkp/pkp-lib/blob/72087f4dabb1e80629b33639e3568501fd309fe2/controllers/grid/settings/user/UserGridRow.php#L84 

Most complex likely is LoginAs. For that I would ask Hafsa, whether she could introduce it to the /users endpoint as well, because she worked on the reviewers - https://github.com/pkp/pkp-lib/issues/10290

But meanwhile check whether the other checks (for other actions) would be easy to implement client side or whether there is also some additional complexity, where you don't have enough information client side and it would be necessary to calculate it server side.

ipula:
> @withanage @ipula Whats missing is to determine which actions would be available to current user. Here is the logic from original table https://github.com/pkp/pkp-lib/blob/72087f4dabb1e80629b33639e3568501fd309fe2/controllers/grid/settings/user/UserGridRow.php#L84
> 
> Most complex likely is LoginAs. For that I would ask Hafsa, whether she could introduce it to the /users endpoint as well, because she worked on the reviewers - [pkp/pkp-lib#10290](https://github.com/pkp/pkp-lib/issues/10290)
> 
> But meanwhile check whether the other checks (for other actions) would be easy to implement client side or whether there is also some additional complexity, where you don't have enough information client side and it would be necessary to calculate it server side.

We can check current user id matching with the userId in the actions in vuejs side and hide the actions that not allowed to by the same user.

ipula:
@jardakotesovec  I added the permission check on the loginAs and mergeUser based on https://github.com/pkp/pkp-lib/pull/10658. Please have a look

src/pages/userInvitation/UserInvitationDetailsFormStep.vue
```diff
@@ -44,6 +44,23 @@
 				:heading="t('user.affiliation')"
 				:value="localize(store.invitationPayload.affiliation)"
 			></FormDisplayItemBasic>
+			<FormDisplayItemBasic
+				v-if="store.invitationPayload.phone"
```
jardakotesovec, line 48: Lets handle this in separate PR. There will be more metadata to add I think anyway.

src/managers/UserAccessManager/UserAccessManagerStore.js
```diff
@@ -0,0 +1,116 @@
...
+
+		/**
+		 * Actions
+		 */
+
+		/*
+		 * redirect to send invitation page
+		 */
+		const {pageUrl: sendInvitationPageUrl} = useUrl('invitation/editUser');
+		function editUser(userObj) {
```
jardakotesovec, line 71: Its a detail, but as convention I always pass it as object, so there can be more data added later on. It happened me multiple times that there was more to add.  Could you adjust it so its  editUser({userObj})? Same for other action function as well.

src/managers/UserAccessManager/UserAccessManagerStore.js
```diff
@@ -0,0 +1,116 @@
...
+			window.location = sendInvitationPageUrl.value + '/' + userObj.id;
```
jardakotesovec, line 72: useUrl provides also redirectToPage, so we have same logic for redirects like this.

So something like this should work:
```
const {redirectToPage} = useUrl(`invitation/editUser/${userObj.id}`);
redirectToPage()
```

src/managers/UserAccessManager/useUserAccessManagerActions.js
```diff
@@ -0,0 +1,195 @@
...
+			component: 'grid.settings.user.UserGridHandler',
+			op: 'edit-disable-user',
+			params: {
+				rowId: user.id,
+				enable: user.disabled ? '1' : '',
+			},
+		});
+
+		openLegacyModal({title: t('grid.user.disable')}, (closeData) => {
+			if (closeData.dataChanged[0]) {
```
jardakotesovec, line 94: Seems unnecessary?

jardakotesovec (changes requested):
This looks great! Basically very minor comments.

Only additional thing I would like to ask you to improve flexibility on extensibility.

Good example is the `getItemActions`, which will make possible to add some actions from plugin.

Using similar pattern and get list of columns from function so can we make possible to also add column from plugins?
Good example is here https://github.com/pkp/ui-library/blob/main/src/managers/GalleyManager/useGalleyManagerConfiguration.js#L59

PR #438, opened 2024-10-29
pkp/pkp-lib#9707 use weblate locales for ui

bozana:
s. https://github.com/pkp/pkp-lib/issues/9707

### DIFF
--- a/src/mixins/localizeMoment.js
+++ b/src/mixins/localizeMoment.js
@@ -11,15 +11,22 @@ export default {
 		 * This only maps the locales that need to be translated.
 		 * Most PKP locales work fine in Moment.js.
 		 *
-		 * @param {String} locale The PKP locale, eg - `sr_RS@cyrillic`
+		 * @param {String} locale The PKP locale, eg - `sr_Cyrl`
 		 */
 		getMomentLocale(locale) {
+			locale.replaceAll('_', '-').toLowerCase();
 			const map = {
-				'sr_RS@latin': 'sr',
-				'sr_RS@cyrillic': 'sr-cyrl',
-				'uz_UZ@latin': 'uz-latn',
+				'bs-latn': 'bs',
+				hy: 'hy-am',
+				'nb-no': 'nb',
+				oc: 'oc-lnc',
+				pa: 'pa-in',
+				'pt-pt': 'pt',
+				'sr-latn': 'sr',
+				tl: 'tl-ph',
+				ug: 'ug-cn',
+				'zh-hans': 'zh-cn',
 			};
-
 			return map[locale] ?? locale;
 		},
 	},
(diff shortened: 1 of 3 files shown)


jonasraoni (approved):

src/mixins/localizeMoment.js
```diff
@@ -11,13 +11,13 @@ export default {
...
+		 * @param {String} locale The PKP locale, eg - `sr_Cyrl`
 		 */
 		getMomentLocale(locale) {
 			const map = {
-				'sr_RS@latin': 'sr',
-				'sr_RS@cyrillic': 'sr-cyrl',
-				'uz_UZ@latin': 'uz-latn',
+				sr_Latn: 'sr',
+				sr_Cyrl: 'sr-cyrl',
+				uz_Latn: 'uz-latn',
```
jonasraoni, line 20: Given the locales are similar, another option would be to map just the required locales (e.g. `sr`) and try to convert the remaining automatically (e.g. `locale.replaceAll('_', '-').toLowerCase()`).

I think nobody will remember to update this part of the code, so it seems to be better to have a conversion that works most of the times, than relying on developers to cover all details 🙃

bozana:
Let me ask @jardakotesovec, if all this all is needed and how -- I could not find where it is used and how... :sweat_smile:

jonasraoni:
There's a sample here:
https://github.com/pkp/ui-library/blob/a67fd26f7fca5db4b5a0cc3ef3ddc330390ffb64/src/mixins/autosave.js#L139

The application locale should match the moment locales available at:
`lib/pkp/lib/weblateLanguages/languages.json`

bozana:
Thanks @jonasraoni!
I also find the list of Moment.js locales -- they seam to be in the folder moment/src/locale/.
Do you maybe know if the JS function getMomentLocale always gets our existing UI locales or could it also be a submission locale (which can be any weblate locale)? If it also gets submission locales than a bigger matching from weblate locales to moment.js locales is eventually needed...

bozana:
And what about those missing in Moment.js ?

bozana:
I took a look at the whole list of Moment.js locales and the related Weblate locales and have added also these:
Weblate -> Moment.js
hy -> hy-am
oc -> oc-lnc
pa -> pa-in
tl -> tl-ph
(there is only one option in Moment.js and I took that)
These I could not match because there are several options in Moment.js, so I do not know which one is correct:
en -> ?
gom -> gom-deva, gom-latn
Any suggestion?
Else they can then stay as they are...

@jonasraoni, OK so?

jardakotesovec:
@bozana afaik the date formatting does not really work..  Last time I checked I shared the finding here.. hoping to get it early next year - https://github.com/pkp/pkp-lib/issues/9733

bozana:
Thanks @jardakotesovec, I will leave this part of the code as is -- as it would work :sweat_smile: -- and then when you address the issue you can see what to do with it all... (it seems we will maybe remove it fully...) :-)

jonasraoni (approved):

PR #439, opened 2024-10-30
pkp/pkp-lib#9658 user invitation minor fixes

ipula:


src/pages/userInvitation/UserInvitationSearchFormStep.vue
```diff
@@ -59,13 +59,13 @@ function updateField(fieldName, b, fieldValue) {
 const searchPhrase = computed(() => {
 	let searchText = '';
 	if (fields.value.email) {
-		searchText = fields.value.email + ' ';
+		searchText = fields.value.email.trim() + ' ';
```
ewhanson, line 62: Is it possible any of these could be `null` or `undefined`? If they're guaranteed to be strings, this looks good to me.

ewhanson (changes requested):
Thanks, @ipula! Just one question, but otherwise looks good.

ewhanson (approved):

ewhanson:
Thanks, @ipula. Merged!

PR #442, opened 2024-11-05
pkp/pkp-lib#10050 [stable-3_4_0] Error handling in SubmissionWizard crashes

blesildaramirez:


jardakotesovec (approved):

src/components/Container/SubmissionWizardPage.vue
```diff
@@ -636,7 +636,7 @@ export default {
 					},
 					error(r) {
 						if (!r.responseJSON) {
-							this.ajaxErrorCallback();
+							this.ajaxErrorCallback({});
```
Vitaliy-1, line 639: @blesildaramirez, just wondering, if there is an error without JSON, do we show any default message? Or is it handled other way?

PR #444, opened 2024-11-06
pkp/pkp-lib#1660 Customizable Reviewer Recommendations

touhidurabir:
for pkp/pkp-lib#1660


src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
```diff
@@ -0,0 +1,470 @@
...
+
+			if (typeof recommendation === 'undefined') {
+				this.ajaxErrorCallback({});
+				return;
+			}
+
+			this.openDialog({
+				name: 'delete',
+				title: this.deleteRecommendationLabel,
+				message: this.replaceLocaleParams(this.confirmDeleteMessage, {
```
asmecher, line 364: This is an XSS flaw -- any HTML entered in the `recommendation.title` will be passed through to the browser. (Also the other keys that mix user-supplied data in here.)

asmecher (changes requested):

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
touhidurabir: @asmecher good spot . I had this left out as `v-strip-unsafe-html` was not yet available for `main` as we introduced in `3.4.0` . But then forgot to update these as `v-strip-unsafe-html` was added just about a week ago . 

One question , rather than applying this for `replaceLocaleParams` every time like 
```js
message: this.replaceLocaleParams(this.confirmDeleteMessage, {
    title: sanitizeHtml(this.localize(recommendation.title)),
})
```
should we add this `sanitizeHtml` into the implementation of `replaceLocaleParams` so that we don't need to make sure to add this every time ?

jardakotesovec: XSS is by default prevented by vue.js, and in cases where we use (and sometime overuse) v-html, its now replaced with v-strip-unsafe-html.  Therefore this will go through that. Every component is responsible to do the XSS protection.

Therefore Dialog handles that when rendering message.
```
<div v-strip-unsafe-html="message" />
```

No point to sanitize twice.

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
```diff
@@ -0,0 +1,472 @@
...
+import TableBody from '@/components/Table/TableBody.vue';
+import TableRow from '@/components/Table/TableRow.vue';
+import TableCell from '@/components/Table/TableCell.vue';
+import PkpButton from '@/components/Button/Button.vue';
+import DropdownActions from '@/components/DropdownActions/DropdownActions.vue';
+import ajaxError from '@/mixins/ajaxError';
+import dialog from '@/mixins/dialog.js';
+import fetch from '@/mixins/fetch';
+import cloneDeep from 'clone-deep';
+import ReviewerRecommendationsEditModal from './ReviewerRecommendationsEditModal.vue';
```
jardakotesovec, line 77: I see that you created some components with options api and some with composition api. This looks bit of mix of both. Mainly options API, but also using composables which are mainly intended for composition API.

Any chance we could have this consistently using composition api.

There is no need to use ajaxError, dialog, fetch mixins. We have all these covered in useModal and useFetch. We already have lots of usage examples and these are actually quite well documented and I am also able to respond on mattermost for any usage questions.

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
```diff
@@ -0,0 +1,472 @@
+<template>
```
jardakotesovec, line 1: I am putting these new 'managing' components to the src/managers

We could name it something like  ReviewerRecommendationsManager

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
```diff
@@ -0,0 +1,472 @@
...
+						<span
+							v-strip-unsafe-html="localize(item.title)"
+							class="text-lg-normal"
+						></span>
+					</TableCell>
+
+					<TableCell>
+						<label>
+							<input
+								class="text-blue-900 bg-gray-100 border-gray-300 focus:ring-blue-500 dark:focus:ring-blue-600 dark:ring-offset-gray-800 dark:bg-gray-700 dark:border-gray-600 h-5 w-5 rounded focus:ring-2"
```
jardakotesovec, line 36: Styled checkboxes probably come with redesigned forms.

So yeah, we don't have that yet. It would be fine to use unstyled checkbox meanwhile. But not big deal either way.

jardakotesovec (changes requested):
Thanks @touhidurabir. This is very nicely integrated with new code base. I am not sure whether this is aimed for 3.5, so I would understand if some of the 'composition api' updates would be left out for timeline purposes.

But if you decide to do the upgrade and struggle with something feel free to message me directly on mattermost. I should be able to point you to some good example or something.

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
asmecher: @jardakotesovec, I used a `<i>` tag to confirm that entered HTML was passed through, and I didn't check whether more malicious tags were filtered -- so it's possible that `<i>` gets through because it's "safe", but worse HTML would be filtered. However, it's still incorrectly escaping the content. The reviewer recommendation title is plaintext, not a rich text field, so any entered content should be fully escaped before getting glued into the locale string.

asmecher:
> I am not sure whether this is aimed for 3.5

It's scheduled for 3.6, and I was considering pulling it forward to 3.5 -- but I'd rather spend a bit more time to get a clean implementation than rush it in. So let's leave it on the 3.6 milestone.

asmecher:
@jardakotesovec, would you be able to review the string escaping stuff here? I'm not sure it's matching what you've been working towards for new code.
@touhidurabir, there's a merge conflict here.

src/components/ListPanel/reviewerRecommendations/ReviewerRecommendationsListPanel.vue
touhidurabir: Modified to use the composition API .

src/managers/ReviewerRecommendationManager/ReviewerRecommendationManager.vue
```diff
@@ -0,0 +1,140 @@
...
+				<TableColumn>{{ recommendationStatusTitle }}</TableColumn>
+				<TableColumn></TableColumn>
+			</TableHeader>
+
+			<TableBody>
+				<TableRow v-for="item in store.items" :key="item.id">
+					<TableCell :is-row-header="true">
+						<span
+							v-strip-unsafe-html="escapeHtml(localize(item.title))"
+							class="text-lg-normal"
```
jardakotesovec, line 29: What behaviour are you aiming here exactly?  Vue.js will escape html by default, so wondering how is this different to just use
```
<span>{{localize(item.title}}</span>
```

src/managers/ReviewerManager/reviewerManagerStore.js
```diff
@@ -36,7 +36,7 @@ export const useReviewerManagerStore = defineComponentStore(
 		/**
 		 * Config
 		 */
-		const reviewerManagerConfig = extender.addFns(useReviewerManagerConfig());
+		const reviewerManagerConfig = extender.addFns(useReviewerManagerConfig(props.recommendations));
```
jardakotesovec, line 40: Just small convention that I aim to follow is to use object so its better flexibility to pass more data if needed, without dealing with the argument order.

So please just adjust it to :
```
const reviewerManagerConfig = extender.addFns(useReviewerManagerConfig({recommendations: props.recommendations}));
```

src/managers/ReviewerRecommendationManager/ReviewerRecommendationManager.vue
```diff
@@ -0,0 +1,140 @@
...
+	id: {type: String, required: true},
+	title: {type: String, required: true},
+});
+
+const {t, localize} = useLocalize();
+const store = useReviewerRecommendationManagerStore(props);
+
+// Actions handling
+function getActions(item) {
+	const actions = [
```
jardakotesovec, line 97: Since you already created store, which is great. Just put all the logic in there.

The motivations why its organised like this are explained in https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-pinia-store--docs and https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-plugins--docs


Long story short putting all logic to the store for easier extensibility. And it would be difficult to draw line what kind of logic to put directly to the component and which to the store. So if the component has its own store - lets just put everything there.

getActions is actually good example of the things that we might want to have extensible and having that exposed in the store, like we do in FileManager/GalleyManager/ParticipantManager/ReviewerManager...  would make sense. But probably unncessary, because I think its very unlikely someone will need to extend this particular UI.

src/managers/ReviewerRecommendationManager/ReviewerRecommendationManager.vue
```diff
@@ -0,0 +1,140 @@
...
+import {escapeHtml} from '@/directives/stripUnsafeHtml';
+
+const props = defineProps({
+	addRecommendationLabel: {type: String, required: true},
+	confirmDeleteMessage: {type: String, required: true},
+	deleteRecommendationLabel: {type: String, required: true},
+	editRecommendationLabel: {type: String, required: true},
+	activateTitle: {type: String, required: true},
+	confirmActivateMessage: {type: String, required: true},
+	deactivateTitle: {type: String, required: true},
```
jardakotesovec, line 82: No need to pass translations as props anymore! https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-translation--docs

Just use useLocalize and t('localisationKey') and it will just work. That will dramatically reduce number of props

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,212 @@
...
+						isPrimary: true,
+						callback: async (close) => {
+							const success = await _actionFns.toggleStatus({
+								id: item.id,
+								newStatus,
+							});
+							if (success) {
+								// Only update the specific item's status
+								items.value.forEach((i) => {
+									if (i.id === item.id) {
```
jardakotesovec, line 61: How I think about the server updates is that ultimately what I am showing to the user should come to the user as that will be most accurate.
Making adjustments to the data client side is bit of additional work, which somewhat duplicates the edits that are happening on the server. The strategy sometime used is to adjust data client side (often called optimistic update) and on background reload the data from the server to ensure accuracy.  

But in our use cases I am usually aiming for simplicity, as these API calls are quick and not often. So basically just send the update to the server and refetch recommendations will be perfectly sufficient.

Therefore you can simplify it to

```js
await _actionFns.toggleStatus({...})
await fetchRecommendations();
```

And error handling also is already solved for you if you use useFetch to display that server returned error.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		function handleAdd() {
+			const activeForm = cloneDeep(props.form);
+			activeForm.action = props.apiUrl;
+			activeForm.method = 'POST';
+
+			// Initialize multilingual fields properly
+			activeForm.fields = activeForm.fields.map((field) => {
+				const baseField = {...field};
```
jardakotesovec, line 126: So there is `useForm` to abstract these manually updates to the form structure. It was created quickly to cover some use cases in 3.5.. it will mature more in 3.6, but I think this would be also good example where to use it:

https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/composables-useform--docs

Please try  const {form, clearForm} = useForm(props.form);
clearForm()

Message me on mattermost if its not working for you, and I will improve it.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,212 @@
...
+					baseField.value = field.value || '';
+				}
+				return baseField;
+			});
+
+			openSideModal(ReviewerRecommendationsEditModal, {
+				title: props.addRecommendationLabel,
+				activeForm,
+				onUpdateForm: (formId, data) => {
+					Object.keys(data).forEach((key) => {
```
jardakotesovec, line 142: When passing form configuration to the modal - I would suggest just to pass the configuration and let the modal handle the state of it.

Than you can also use the useForm() again for managing state of that form - https://github.com/pkp/ui-library/blob/d0d5dfb9389e8d3b80ef6985eefd01dfa4b41aaa/src/managers/ReviewerManager/modals/workflowLogResponseModalStore.js#L4

You can see that in this example Rob created own store for the modal, which is fine. But for simple modals, where we don't expect devs to extend/amend the behaviour, its fine to put that logic directly to the component.

Its intentional that the syntax in the component/store is exactly the same, so its easy to move things to the store if needed.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,212 @@
...
+						activeForm[key] = data[key];
+					});
+				},
+				onFormSuccess: (result) => {
+					items.value = [...items.value, result];
+					pkp.eventBus.$emit('add:recommendation', result);
```
jardakotesovec, line 148: At this point I intend to handle extensibility just via stores, and avoid these emits, unless we find reason this to be necessary. So this can be removed.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,212 @@
...
+
+			openDialog({
+				name: 'edit',
+				title: newStatus ? props.activateTitle : props.deactivateTitle,
+				message: replaceLocaleParams(
+					item.status
+						? props.confirmDeactivateMessage
+						: props.confirmActivateMessage,
+					{
+						title: escapeHtml(localize(item.title)),
```
jardakotesovec, line 46: I guess that the reviewer recommendations are not supporting any html tags, right? So we basically just want to make sure that if it has any html tags it would not be interpreted?

We have couple places like tooltip or dialogs, where we introduced v-html because we had some message in translations which required that. To reduce attack surface area we introduced v-strip-unsafe-html, which at least ensures safe html.

I would like to improve the situation in 3.6 further.  

I think for general components that because of some use case started offering v-html - we should have separate props for that. For Example having openDialog with options: `message` and `messageHtml`. So its decided when the dialog is created whether require html being rendered and whether its safe based on what I am displaying. For this improvement I will create separate ticket and introduce the `html` variants when necessary.

But your use case is bit more nuanced, because you have translation `Are you sure you want to deactivate the recommendation <b>{$title}</b>`. Therefore you want to render it as html, but escape the argument, which you are doing, but would be nice to also make the translation toolkit safe by default.

I will leave up to you whether you would add this improvement here, since you already touched on the escaping functions or whether you prefer me to add it in separate ticket, but I think having option in our `t` function will be best..

By default `t` should escape all the arguments being passed. But need to introduce option to allow html if needed.

So in your case it should look like (now putting aside that I will move the message to messageHtml in separate ticket):

```
message: item.status ? t('manager.reviewerRecommendations.confirmDeactivate', {title: localize(item.title)}) : ...
```

But there still would be option, to skip the escaping for specific fields:
t('manager.reviewerRecommendations.confirmDeactivate', {title: localize(item.title)}, {allowHtml: ['title']}) 



What that means for this work is that you can go with

src/managers/ReviewerRecommendationManager/ReviewerRecommendationManager.vue
touhidurabir: Right, I don't need to use the `v-strip-unsafe-html` here .

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
touhidurabir: I will and want to work on this but I think it will be better to handle this as a separate issue . Reason is, this would be a global impact level change and a very important one . Better to keep track of this as it's own issue so that we can in future get clear idea what plan and strategy we adapted for this one and what was the implementation. This issue is already got bit too big so better to separate this implementation from this issue .

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,205 @@
...
+								? item[field.name]
+								: field.value || '';
+					}
+
+					return baseField;
+				});
+			}
+
+			currentForm.value = preparedForm;
+		}
```
touhidurabir, line 161: seems at edit time, the form does not prefill wilth existing data and there is consideration for multilingual and non-multilingual data . So I used this approach but not sure if this is proper approach .

touhidurabir:
Hi @jardakotesovec can you please take another look .

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
jardakotesovec: Agree.. make sense to introduce these changes separately. 

In that case I would suggest not to introduce the `escapeHtml` at this point. Because that will be default behaviour for the `message` of dialog.  So if you remove it - it will be safe enough for now and after we work on the updates that I mentioned, it will also prevent sneaking in any html to affect rendering.

jardakotesovec (changes requested):
Regarding useForm, sorry I need to document this better. I did some testing and it seems to be working fine. Feel free to cherry-pick this commit - https://github.com/jardakotesovec/ui-library/commit/6e4ec1a7572969283ac5ee80e98673e3ad3bbc27

I also sneaked in spinner to the table to indicate when the data are being refetched for small connections.

So my suggestion would be to add this commit (seemed working fine from my testing, if you see issue ping me on mattermost).

Resolve other suggestions.

And ping me and I will do the last quick check.

Thank you!

touhidurabir:
@jardakotesovec hopefully ready for the final look , please check .

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,169 @@
...
+		async function toggleStatus({id, newStatus}) {
+			const {isSuccess, fetch} = useFetch(`${apiUrl.value}/${id}/status`, {
+				method: 'PUT',
+				body: {
+					status: Number(newStatus),
+				},
+			});
+
+			await fetch();
+			return isSuccess.value;
```
jardakotesovec, line 38: Minor, but isSuccess is not really needed.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		async function handleStatusToggle(item, event) {
+			const newStatus = !item.status;
+
+			openDialog({
+				name: 'edit',
```
jardakotesovec, line 54: minor, name is not used anymore

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,169 @@
...
+				title: newStatus
+					? t('manager.reviewerRecommendations.activate.title')
+					: t('manager.reviewerRecommendations.deactivate.title'),
+				message: t(
+					item.status
+						? 'manager.reviewerRecommendations.confirmDeactivate'
```
jardakotesovec, line 60: Alright.. :-).. this is tricky..

Automatic detection of the locale keys is happening in build time and therefore I have to rely on regex.

So it needs to find t('key') or tk('key') to pick it up.

With having condition like this inside, I don't think it would work.

src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js
```diff
@@ -0,0 +1,169 @@
...
+			setMethod('PUT');
+			setValues(item);
+			openRecommendationFormModal(
+				t('grid.action.editReviewerRecommendation'),
+				editForm,
+			);
+		}
+
+		// Initial data fetch
+		fetchRecommendations();
```
jardakotesovec, line 156: This is very subjective suggestion about how to organise with composition API. Especially for more complex components trying to group things by feature.  So since the beginning is focused on the fetching the recommendations I would do the initial trigger there.

```
		const {
			data: recommendations,
			fetch: fetchRecommendations,
			isLoading: isRecommendationsLoading,
		} = useFetch(apiUrl);
		// Initial data fetch
		fetchRecommendations();
```

jardakotesovec (changes requested):
Just couple small clean up items. Thanks!

PR #449, opened 2024-11-20
pkp/pkp-lib#10624 Add customization support for FieldBaseAutosuggest options

blesildaramirez:


src/components/Form/fields/Autosuggest.vue
```diff
@@ -0,0 +1,164 @@
...
+
+const slots = useSlots();
+
+const props = defineProps({
+	id: {
+		type: String,
+		required: true,
+	},
+	inputProps: {
+		type: Object,
```
jardakotesovec, line 98: Is inputProps needed? For what exactly?

src/components/Form/fields/FieldRorAutosuggest.vue
```diff
@@ -0,0 +1,154 @@
...
+const inputValue = ref('');
+const isFocused = ref(false);
+
+const url = 'https://api.ror.org/v2/organizations';
+const queryParams = ref({
+	query: inputValue.value,
+});
+const {data, isLoading, fetch} = useFetch(url, {
+	query: queryParams,
+});
```
jardakotesovec, line 69: const {data: suggestions, isLoading, fetch: fetchSuggestions} = useFetch(url, {
	query: queryParams,
});

const queryParams = computed(() => ({query: inputValue.value}))

watch(queryParams, () => { fetchSuggestions() })

const mappedSuggestions = computed(() => suggestions.value.map(...))

PR #460, opened 2024-12-10
pkp/pkp-lib#9661 Allow Journal Managers to invite users to adopt a role - ORCiD

ewhanson:


src/pages/acceptInvitation/AcceptInvitationPageStore.js
```diff
@@ -311,7 +398,12 @@ export const useAcceptInvitationPageStore = defineComponentStore(
 				method: 'PUT',
 				body: {invitationData: invitationRequestPayload.value},
 			});
-			if (!acceptInvitationPayload.value.privacyStatement) {
+			// FIXME: Privacy statement check blocks ORCID from saving before hand.
+			//  		  Can this check be moved outside of update payload check?
```
bozana, line 402: Who could best help with this question?

ewhanson: Probably @jardakotesovec, but I know he's quite busy. If he and you think it's okay, we could merge this as-is, and I could tackle this as a separate issue to keep things moving. What do you think?

PR #471, opened 2024-12-20
pkp/pkp-lib#10624 Allow custom value when selecting affiliation

blesildaramirez:


src/components/Form/fields/FieldRorAutosuggest.vue
```diff
@@ -70,6 +70,8 @@ const queryParams = computed(() => ({
 	query: inputValue.value,
 }));
 
+provide('allowCustom', true);
+
```
jardakotesovec, line 74: Hello,
just wondering why this would be communicate via provide and not normal prop?

PR #493, opened 2025-01-22
pkp/pkp-lib#1550 Controlled vocabulary support

jyhein:
- Use external source to add new keywords etc. using plugins
- Save extra data (identifier, source) with the term to db

### DIFF
--- a/src/components/Form/fields/Autosuggest.vue
+++ b/src/components/Form/fields/Autosuggest.vue
@@ -95,7 +95,25 @@
 							name="option"
 							:suggestion="suggestion"
 						/>
-						<span v-else>{{ suggestion.label }}</span>
+						<ul v-else>
+							<li>
+								{{ suggestion.label }}
+							</li>
+							<li v-if="suggestion.identifier?.match(/^http/)">
+								<a :href="suggestion.identifier" target="_blank" @click.stop>
+									{{ suggestion.identifier }}
+								</a>
+							</li>
+							<li v-else-if="suggestion.identifier">
+								{{ suggestion.identifier }}
+							</li>
+							<li
+								v-for="(extraItem, extraItemKey) in suggestion.extraItems ?? {}"
+								:key="extraItemKey"
+							>
+								{{ extraItem }}
+							</li>
+						</ul>
 					</li>
 				</ComboboxOption>
 			</template>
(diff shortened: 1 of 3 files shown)


src/components/Form/fields/FieldControlledVocab.vue
```diff
@@ -2,6 +2,17 @@
...
+	const {name, label = null, identifier = null, ...extraItems} = value;
+	const suggestion = {
+		value: value,
+		label: label ?? name,
+		identifier: identifier,
+		...(extraItems ? {extraItems: extraItems} : {}),
+	};
+	return suggestion;
+}
+
```
touhidurabir, line 15: Any particular reason we are making this as general method than as the part of component's method ? Also If I understand correctly, this method takes the data coming from plugin and transform it to support the updated schema . If that is the case , perhaps a different name so that it does not confuse the `setSuggestions` method coming form `FieldBaseAutosuggest` component ?

src/components/Form/fields/Autosuggest.vue
```diff
@@ -95,7 +95,25 @@
...
+								{{ suggestion.label }}
+							</li>
+							<li v-if="suggestion.identifier?.match(/^http/)">
+								<a :href="suggestion.identifier" target="_blank" @click.stop>
+									{{ suggestion.identifier }}
+								</a>
+							</li>
+							<li v-else-if="suggestion.identifier">
+								{{ suggestion.identifier }}
+							</li>
```
touhidurabir, line 109: Can we merge these 2 together as first we need to check if `identifier` exists and then if it is an url , then wrap it with anchor tag or just show it as plain text ?

src/pages/workflow/components/publication/WorkflowPublicationForm.vue
```diff
@@ -55,7 +55,16 @@ const displayNoFieldsEnabled = computed(() => {
...
+	metadata: metadataDataChange,
+};
+
+const {triggerDataChange} = useDataChanged(customFns[props.formName]);
 
 const {set, form} = useForm(publicationForm);
+
+async function metadataDataChange() {
+	// Some metadata fields need extra data from db not in publication object
+	fetchForm();
```
touhidurabir, line 68: should there be a `await` here ?

touhidurabir (commented):
@jyhein I added few questions/suggestion, nothing critical . Please take a look when you get time .

src/components/Form/fields/FieldControlledVocab.vue
jyhein: Hmmm...shouldn't this be private? When it is outside export

PR #494, opened 2025-01-25
pkp/pkp-lib#10820 Display error summary at the top of the page with multiple forms

blesildaramirez:


jardakotesovec (changes requested):
Looking good.

Maybe check the messaging with Devika, as the wording **3 errors detected! Please correct the following errors before proceeding.** works well if the errors would be listed. But since we don't listen to them.. maybe something like 'correct errors below before proceeding' would be better fit?

Once thats settled you can add translation to the common.po

jardakotesovec (approved):

PR #495, opened 2025-01-25
pkp/pkp-lib#10842 Extend DropdownActions component with text button display option

blesildaramirez:


jardakotesovec (approved):
Looking good, only suggestion I would have to use word 'variant' instead of type. As thats what we often use to differentiate different stylings.  `buttonVariant`.

And chromatic deployment is failing.. can you check?

PR #515, opened 2025-02-06
pkp/pkp-lib#9658 changes in invitation and user access table

ipula:
see https://github.com/pkp/pkp-lib/issues/10916

### DIFF
--- a/src/pages/userInvitation/UserInvitationPageStore.js
+++ b/src/pages/userInvitation/UserInvitationPageStore.js
@@ -301,7 +301,11 @@ export const useUserInvitationPageStore = defineComponentStore(
 		}
 
 		const {redirectToPage} = useUrl('management/settings/access');
-		const isSubmitting = ref(invitationPayload.value.disabled);
+		const isSubmitting = ref(
+			invitationMode.value === 'editUser'
+				? true
+				: invitationPayload.value.disabled,
+		);
 
 		/**
 		 * change isSubmitting value based on
@@ -311,6 +315,9 @@ export const useUserInvitationPageStore = defineComponentStore(
 			invitationPayload,
 			async (newVal, oldVal) => {
 				isSubmitting.value = invitationPayload.value.disabled;
+				if (invitationPayload.value.userGroupsToAdd.length === 0) {
+					isSubmitting.value = true;
+				}
 				detectChanges.value = true;
 			},
 			{deep: true},


src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -267,7 +267,7 @@ export const useUserInvitationPageStore = defineComponentStore(
 			await createInvitation();
 
 			if (validationError.value) {
-				errors.value = validationError.value.errors;
+				errors.value = validationError.value.errors['inviteeEmail'];
```
blesildaramirez, line 270: Hi @ipula , I checked this PR and tested on my local. I believe for this line, you intend to remove the "userId" from the list of errors coming from the API. The current code made the errors prop like this:
```
['This field is required when user id is not present.']
```
To display errors, the component expects the field id in the object in this format:
```
{
  inviteeEmail: ['This field is required when user id is not present.']
}
```

To resolve this, we can change this to:
```suggestion
                                delete validationError.value.errors['userId'];
				errors.value = validationError.value.errors;
```

blesildaramirez (changes requested):

src/pages/userInvitation/UserInvitationPageStore.js
ipula: yes, my mistake I will fix this

src/pages/userInvitation/mocks/pageInitConfig.js
```diff
@@ -62,7 +62,7 @@ export default {
 			name: 'Search User',
 			stepLabel: '{$step} - Search User',
 			description:
-				'Search for the user using their email address, username or ORCID iD. Enter at least one details to get started. If user does not exist, ypu can invite them to take up roles and be a part of your journal. If the user already exist in the system, you can view user information and invite to take a additional roles.',
+				'If user does not exist, ypu can invite them to take up roles and be a part of your journal. If the user already exist in the system, you can view user information and invite to take a additional roles.',
```
blesildaramirez, line 65: Sorry I missed this on my first review. There's a small typo here and message could be improved:
```suggestion
				'If the user does not exist, you can invite them to take on roles and become part of your journal. If the user already exists in the system, you can view their information and invite them to take on additional roles.',
```

blesildaramirez (changes requested):

public/globals.js
```diff
@@ -855,7 +856,19 @@ window.pkp = {
...
 		'workflow.uploadRevisions': 'Upload revisions',
-			'userInvitation.searchField':'Search for a user by email address, username, or ORCID iD. Enter only one to get started!'
+		'userInvitation.searchField':
+			'Search for a user by email address, username, or ORCID iD. Enter only one to get started!',
+		'grid.user.enable': 'Enable User',
+		'grid.user.disable': 'Disable User',
+		'grid.user.remove': 'Remove User',
+		'grid.user.logInAs': 'Login As',
+		'grid.user.grid.user.enableReasonDescription':
+			"Once the user enabled, they will regain access to OJS, and you'll be able to invite them to roles as needed.",
```
blesildaramirez, line 866: Looks like "is" is missing here:
```suggestion
			"Once the user is enabled, they will regain access to OJS, and you'll be able to invite them to roles as needed.",
```

src/managers/UserAccessManager/useUserAccessManagerActions.js
```diff
@@ -91,10 +89,24 @@ export function useUserAccessManagerActions() {
 				enable: user.disabled ? '1' : '',
 			},
 		});
-
-		openLegacyModal({title: t('grid.user.disable')}, (closeData) => {
-			finishedCallback();
+		let currentRoles = '';
+		user.groups.forEach((group) => {
+			currentRoles = currentRoles + group.name + ',';
 		});
```
blesildaramirez, line 95: This can be changed to:
```suggestion
		const currentRoles = user.groups.map(group => group.name).join(', ');
```
Note that space might be needed for a more structured message on the modal, when passing the description value to `openLegacyModal` function call below.

src/managers/UserAccessManager/useUserAccessManagerActions.js
```diff
@@ -91,10 +89,24 @@ export function useUserAccessManagerActions() {
...
+
+		openLegacyModal(
+			{
+				title: !user.disabled
+					? t('user.disabledModal.title', {fullName: user.fullName})
+					: t('user.enabledModal.title', {fullName: user.fullName}),
+				description: t('user.disabledModal.description', {
+					roles: currentRoles.replace(/,\s*$/, ''),
```
blesildaramirez, line 103: Given suggestion above, we can remove the need to use regex and replacing string:
```suggestion
					roles: currentRoles,
```

src/pages/userInvitation/UserInvitationPageStore.js
```diff
@@ -300,7 +301,20 @@ export const useUserInvitationPageStore = defineComponentStore(
 		}
 
 		const {redirectToPage} = useUrl('management/settings/access');
-		const isSubmitting = ref(false);
+		const isSubmitting = ref(invitationPayload.value.disabled ? true : false);
```
blesildaramirez, line 304:
```suggestion
		const isSubmitting = ref(!!invitationPayload.value.disabled);
```
or if value is always a boolean for `invitationPayload.value.disabled`, we can just reference it directly:
```suggestion
		const isSubmitting = ref(invitationPayload.value.disabled);
```

blesildaramirez (changes requested):
Just a few code optimization and wording improvements.

src/pages/userInvitation/UserInvitationPageStore.js
ipula: Yes I saw this one

blesildaramirez (approved):

PR #525, opened 2025-02-13
pkp/pkp-lib#10337 Handle 403&404 code in useFetch

blesildaramirez:


jardakotesovec (approved):
Looks great! No objections here :-D.

PR #550, opened 2025-03-04
Allow arbitrary nesting depth in categories

taslangraham:
For pkp/pkp-lib#10449


src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+		const expanded = computed(() => Array.from(_expandedIds.value));
+
+		const categories = computed(() => _categories.value);
+		const currentCategoryForm = ref({});
+		const categoryForm = props.categoryForm; // Blank form used only for adding new categories
+
+		// Store categories at the start of ordering. This will allow to reset data if user cancels ordering.
+		const _categoriesBeforeReordering = ref([]);
+		// IDs of expanded categories will be cleared at the start of ordering to collapse all expanded items.
+		// This will store the IDs of expanded categories at the start of ordering to allow re-expansion after ordering is completed/cancled.
```
ewhanson, line 41: Typo in `canceled`

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+		const _expandedIdsBeforeOrdering = ref(new Set());
+
+		/**
+		 * Prepares the form data for Adding/Editing a Category
+		 * @param {Object} category - Category object which is being edited or a sub category is being added to.
```
ewhanson, line 46: The last part is a little confusing to me. Does it mean when a sub category is having another category added to it (i.e. the subcategory is the parent for a new sub-sub category), `category` will be that subcategory?

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+		 * This wil be null when adding a top level Category.
```
ewhanson, line 47: If it can be null, should the '@param' annotation be `{Object|null}`?

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+			if (Object.entries(category).length && action === Actions.CATEGORY_EDIT) {
+				preparedForm = await getEditForm(category.id);
+				preparedForm.method = 'PUT';
+				preparedForm.action = `${preparedForm.action}/${category.id}`;
+			}
+			currentCategoryForm.value = preparedForm;
+		}
+
+		/**
+		 * Get configured form to edit a category.
```
ewhanson, line 76: Could you include the argument and return type info in the JS doc block here?

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+		 */
+		async function getEditForm(categoryId) {
+			const url =
+				apiUrl.value + `/categoryFormComponent?categoryId=${categoryId}`;
+			const {fetch, data} = useFetch(url, {method: 'GET'});
+			await fetch();
+			return data.value;
+		}
+
+		function getItemActions() {
```
ewhanson, line 86: Missing function description.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+				{
+					label: t('common.delete'),
+					icon: 'Cancel',
+					name: Actions.CATEGORY_DELETE,
+					isWarnable: true,
+				},
+			];
+		}
+
+		async function handleItemAction(actionName, category = null) {
```
ewhanson, line 107: Missing function description, argument JS doc annotations.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+			) {
+				await setupCategoryForm(category, actionName);
+				openCategoryFormModal(ActionFormLabel[actionName]);
+			}
+		}
+
+		/***
+		 * Deletes a category.
+		 */
+		function deleteCategory(category) {
```
ewhanson, line 122: Missing `category` JS doc annotation.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+			const {openDialog, openDialogNetworkError} = useModal();
+			const {apiUrl} = useUrl(`categories/${category.id}`);
+			const {fetch: fetchDelete, isSuccess} = useFetch(apiUrl, {
```
ewhanson, line 125: Does `apiUrl` need to be `apiUrl.value` here since `useFetch `is expecting a `string` and `apiUrl` is a `ComputedRef<string>`? Looks like it's used somewhat inconsistently throughout the codebase so probably fine, but thought I would flag it just in case. Do you have a preference/answer @jardakotesovec?

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+				title: t('common.delete'),
+				message: t('common.confirmDelete'),
+				modalStyle: 'negative',
+			});
+		}
+
+		/**
+		 * Open the form modal to add/edit category.
+		 */
+		function openCategoryFormModal(title) {
```
ewhanson, line 160: Missing argument JS doc type annotation

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+				title,
+				formProps: form,
+				onCategorySaved: categorySaved,
+			});
+		}
+
+		/**
+		 * Function to execute when a category is successfully saved via form.
+		 */
+		async function categorySaved(category) {
```
ewhanson, line 173: Missing argument JS doc block type.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+			isLoading.value = true;
+			const {data, fetch} = useFetch(apiUrl, {method: 'GET'});
+			await fetch();
+			_categories.value = data.value;
+			isLoading.value = false;
+		}
+
+		/**
+		 * Expand or collapse a parent in the category tree.
+		 * @param id - The ID of the category to expand/collapse.
```
ewhanson, line 205: Include type for `id`

src/managers/CategoryManager/CategoryManager.vue
```diff
@@ -0,0 +1,238 @@
...
+								<div class="bg-blue-100 py-1 pl-2">
+									{{ item.value.title[primaryLocale] }}
+								</div>
+							</TreeItem>
+						</TableCell>
+
+						<TableCell>
+							<span class="category__manager__assigendEditors">
+								{{
+									item.value.assignedEditors
```
ewhanson, line 96: Pretty sure this has come up before, but I can't remember the answer. Do we have ny methods for handling localized/internationalized serial objects like this? Or is this sufficient for this use case? I'm also wondering if this would cause issues with RTL languages.

ewhanson (changes requested):
Thanks @taslangraham. I mostly looked a the code from a documentation perspective so I'll let @jardakotesovec give his more specific feedback on the implementation.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,315 @@
...
+
+		/**
+		 * Prepares the form data for Adding/Editing a Category
+		 * @param {Object} category - Category object which is being edited or a sub category is being added to.
+		 * This wil be null when adding a top level Category.
+		 * @param {string} action - Action being performed. One of Actions.CATEGORY_ADD or Actions.CATEGORY_EDIT.
+		 */
+		async function setupCategoryForm(category, action) {
+			// Get a deep copy of the form to eliminate references
+			let preparedForm = JSON.parse(JSON.stringify(categoryForm));
```
jonasraoni, line 52: I think you can use the `structuredClone()`

jonasraoni (approved):
Just skimmed through the code.

src/managers/CategoryManager/categoryManagerStore.js
jardakotesovec: Pretty much everything which might be changing - url, query, body can be ref, so the update gets picked up by useFetch..  Documentation could be better on my part..  maybe one day we will use typescript to make these things more obvious..

jardakotesovec:
@taslangraham On initial check I was curious whether the reka-ui will solve the accessibility part, and I don't think it does. Its because it only offers the role=tree, and from bit more research I learnt that our use case is covered by role=treegrid.

 Good news is that its pretty much table with additional attributes and some keyboard support, which we could probably build in to the Table component itself. Bad news is that we might need to do ourselves at some point. But would be handled in separate issue.

So at this point I would suggest to remove the reka component, unless you think its actually doing something useful :-).

And than ping me when the implementation is updated with the latest changes. Overall I think your approach is great..  so will just check for details at the end.

src/managers/CategoryManager/CategoryManager.vue
jardakotesovec: Interesting that for examples in our publication endpoint, we have
```
    "authorsString": "Rosanna Rossi (Author)",
    "authorsStringIncludeInBrowse": "Rosanna Rossi (Author)",
    "authorsStringShort": "Rossi",

```
which is not multilingual..

I guess its bit tricky to combine multiple multilingual fields..    because there is no guarantee to have names for all languages across all authors/editors. But with some fallbacking could be done.

Maybe @Devika008  could provide some thoughts around whether that would be worthwhile effort.

And I hope that I understood what you are referring to :-)

src/managers/CategoryManager/categoryManagerStore.js
jardakotesovec: I am personally slowly moving to generating jsdocs via claude code or other llm.. its often more accurate than me..

jardakotesovec: Will ask it to refine the useFetch jsdocs ;-)

src/managers/CategoryManager/categoryManagerStore.js
jardakotesovec: To be honest - I much more care about descriptive names and variables.. than having jsdocs which says that it returns actions items :-).

I am than happy if there are comments around more complex code which is not obvious on first sight..   But for most of the time well named things are good enough

src/managers/CategoryManager/categoryManagerStore.js
taslangraham: If the category object exist, then then it represents one of two things:
1. An existing category whose data(title, description, etc.) is being updated. Or
2. An existing category to which we we are to adding a sub category

src/managers/CategoryManager/categoryManagerStore.js
taslangraham: Not sure why, but using `structuredClone()` throws "```Uncaught DataCloneError: Failed to execute 'structuredClone' on 'Window': #<Object> could not be cloned.```"

jonasraoni: I don't know what's inside the category form, but some things are not cloneable, so I'll mark this as resolved.

taslangraham:
@jardakotesovec  @jardakotesovec Thanks for the review. I've made updates based on the feedback given. @jardakotesovec  I dropped the usage of Reka-ui,  as you suggested, and introduced a simple component that uses the `TableRow` component to to display the categories. Please take a look

src/managers/CategoryManager/CategoryManager.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+	<div class="category__manager">
+		<div class="mt-2">
+			<div class="flex justify-between">
+				<div class="flex flex-row items-center gap-x-3 text-2xl-bold">
```
jardakotesovec, line 5: Lets use the table slots to get this layout accurate.

You can check out `GalleyManager.vue` and use the `label` and `top-controls` slots.

And if Devika still prefers the look without the borders for the label&top-controls, we can introduce props for that on Table.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,275 @@
...
+	'categoryManagerStore',
+	(props) => {
+		const {apiUrl} = useUrl('categories');
+		const {t} = useLocalize();
+
+		const _categories = ref(props.categories || []);
+		const _expandedIds = ref(new Set());
+		const isLoading = ref(false);
+
+		const columns = computed(() => props.columns);
```
jardakotesovec, line 31: We have configurable columns in new Managers (reviewer, galley, file, user). Its bit more work to do them configurable so its actually useful. Basically being able to have custom column component to render something new.   For the managers that are unlikely being extended I don't insist having that capability there. But for categories, I can imagine reasons to extend it from the plugin.

Can you follow the pattern that you can find in any of the managers mentioned above, for example in ReviewerManager. There is specifically useReviewerManagerConfig.js composable, which have all the configurable functions. 

Important is to have every column as separate component. Than its possible to define what component to render in that configuration. Therefore in plugin its possible to create&register own column component and add it there.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,275 @@
...
+
+			// return empty form if adding a new top-level category
+			if (!Object.entries(category).length || action === Actions.CATEGORY_ADD) {
+				currentCategoryForm.value = preparedForm;
+				return;
+			}
+
+			// Prepare form for editing existing category
+			if (Object.entries(category).length && action === Actions.CATEGORY_EDIT) {
+				preparedForm = await getEditForm(category.id);
```
jardakotesovec, line 64: The pattern we use, until we can create forms directly in JS is just to pass the form configuration on page load and than set the values/method/url on client side.

I know that you were working on exposing the forms for publications and you might be wondering why we did not set the values on client side as well.  Reason was that there was quite a lot of business logic already implemented in the PHP and moving that to client side would be additional regression risk.

Good example how to interact with the form configurations using `useForm` is for example here https://github.com/pkp/ui-library/blob/main/src/managers/ReviewerRecommendationManager/reviewerRecommendationManagerStore.js#L145

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,275 @@
...
+			const {fetch, data} = useFetch(url, {method: 'GET'});
+			await fetch();
+			return data.value;
+		}
+
+		/**
+		 * Get the actions available for a category item.
+		 * @return {Array<{ label: string, icon: string, name: string, isWarnable?: boolean}>} - The list of actions available for the category item.
+		 */
+		function getItemActions() {
```
jardakotesovec, line 88: This could be also moved to the useCategoryManagerConfig.js as configuration function, as mentioned above.

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,275 @@
...
+			const {closeSideModal} = useModal();
+			closeSideModal(EditCategoryFormModal);
+		}
+
+		/**
+		 * Loads updated list of categories from API.
+		 */
+		async function refreshCategories() {
+			isLoading.value = true;
+			const {data, fetch} = useFetch(apiUrl, {method: 'GET'});
```
jardakotesovec, line 247: This works fine, and still make sense to use this pattern for POST/PUTS.

For GET, when retrieving data. useFetch is already providing useful reactive data, which can be used directly.

So pattern I usually go for directly in the setup function:
```
const {data: categories, fetch: fetchCategories, isLoading} = useFetch(apiUrl);
```

And you don't need to manage this process manually and you can just call the `fetchCategories` when you need to refresh it.

src/managers/CategoryManager/CategoryTreeRow.vue
```diff
@@ -0,0 +1,90 @@
...
+
+		<TableCell>
+			<span class="category__manager__assigendEditors">
+				{{ category.assignedEditors.map((editor) => editor.name).join(', ') }}
+			</span>
+		</TableCell>
+
+		<TableCell>
+			<div
+				class="category__manager__table_row_actions flex cursor-pointer justify-end gap-4"
```
jardakotesovec, line 21: I believe all these `category__` classes are unnecessary.

src/managers/CategoryManager/CategoryTreeRow.vue
```diff
@@ -0,0 +1,90 @@
...
+		</TableCell>
+	</TableRow>
+
+	<template
+		v-if="
+			category.subCategories?.length &&
+			categoryManagerStore.expanded.includes(category.id)
+		"
+	>
+		<CategoryTreeRow
```
jardakotesovec, line 55: This looks like nice pattern. Not sure yet how exactly we will deal with the accessibility part.. that might affect also how we want to structure things..  but because I don't have clear idea about it right now, lets worry about that later.

package-lock.json
jardakotesovec, line 1: Is there reason for this one? That usually is in addition to changes in package.json, which I don't see being present. So maybe this could be discarded.

jardakotesovec (changes requested):
Feel free to reach out, if any of the feedback is not clear, thanks!

src/managers/CategoryManager/categoryManagerStore.js
taslangraham: I had decided to populate the form on server side for a similar reason--specifically this logic that conditionally adds in `FieldOptions` fields

https://github.com/pkp/pkp-lib/blob/bff00e1426e25d609dbb02aed3a9fa9432a7b718/classes/components/forms/context/CategoryForm.php#L139-L189

So `setValues` wouldn't quite work for those fields when populating the form and I wasn't sure if dynamically building out those `FieldOptions` fields on the frontend was a good idea.

jardakotesovec: I see, that makes sense. 

What I am trying to encourage for is based on the guidance from @asmecher, where we want to evolve our API capabilities, and make frontend interacting mainly with the APIs. That also makes the API more useful to 3rd party.

If I break it down to explore the options.

**1) Getting list of editors for FieldOptions**
-  it would require to improve /users endpoint, to filter by roleId and assignedStage (via user group)

Thats how we would do it once we create the Forms purely client side, but given that you already have good implementation and we generate the form configuration server side, I think its reasonable to stick with this approach.


**2) Getting list of editors that are already assigned to given category, to know which one to select**

Second part is setting the values of the Form, when editing Category.

If its feasible to add the userGroupId to the `category.assignedEditors`, than setting the values should be quite straightforward on client side. Information about userGroup is quite useful in general, as it says how the editor is associated with the category.

useForm.js doc is behind, but there is useForm.test.js to see bunch of examples.

It would look something like this (consider it as pseudo code which I have not tested):
```
const {setValues} = useForm(categoryForm);

// just to reshape it, to follow field names: subeditors[userGroupId]:
function transformEditors(assignedEditors) {
  return assignedEditors.reduce((result, editor) => {
    const key = `subEditors[${editor.userGroupId}]`;
    result[key] = result[key] || [];
    result[key].push(editor.id);
    return result;
  }, {});
}

// setValues set values only for field names that do exist, therefore its fine to pass whole category object to keep things simple
setValues({
...category,
...transformEditors(category.assignedEditors)
})

```

taslangraham: @jardakotesovec Thanks for the explanation! I'm going to go ahead and make the changes to have the form fields populated on the client.

src/managers/CategoryManager/CategoryCellExpandControl.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+	<TableCell>
+		<div
+			class="flex justify-center"
+			:style="{'padding-right': `${depth + 0.5}rem`}"
```
jardakotesovec, line 5: padding-right is not friendly with rtl

You can use `padding-inline-start`, which will be left/right based whether its left to right or right to left language.

jardakotesovec: In practice it mostly means that if you need one sided margin/padding you just use the ms/me instead of ml/mr. And same for padding ps/pe instead of pl/pr.

src/managers/CategoryManager/CategoryManagerCellAssignedTo.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+	<TableCell>
+		<span>
+			{{ category.assignedEditors.map((editor) => editor.name).join(', ') }}
```
jardakotesovec, line 4: We even have localised comma, so its more flexible for some language which would use different delimiter:

you can just use ...join(t('common.commaListSeparator'))

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,273 @@
...
+export const Actions = {
+	CATEGORY_EDIT: 'categoryEdit',
+	CATEGORY_ADD: 'categoryAdd',
+	CATEGORY_DELETE: 'categoryDelete',
+};
+
+const ActionFormLabel = {
+	[Actions.CATEGORY_EDIT]: 'Edit category',
+	[Actions.CATEGORY_ADD]: 'Add category',
+};
```
jardakotesovec, line 23: these needs to be localised..

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,273 @@
...
+		function getItemActions() {
+			return categoryManagerConfig.getItemActions();
+		}
+
+		/*
+		 * Handle the selected action for a category item.
+		 * @param {string} actionName - The name of the action to perform.
+		 * @param {Object|null} category - The category object to perform the action on. This will be null when adding a top-level category.
+		 */
+		async function handleItemAction(actionName, category = null) {
```
jardakotesovec, line 122: Not big deal.. this approach is also perfectly fine. 

Just as convention that I am aiming for is to expose the actions directly on the store. 

so it ends up with  store.categoryAdd, store.categoryEdit  instead of store.handleItemAction

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,273 @@
...
+
+		// const _categories = ref(props.categories || []);
+		const _expandedIds = ref(new Set());
+		const isLoading = ref(false);
+
+		const columns = computed(() => categoryManagerConfig.getColumns());
+		const expanded = computed(() => Array.from(_expandedIds.value));
+		const categories = computed(() => data?.value || []);
+		const categoryForm = props.categoryForm;
+		console.log({props});
```
jardakotesovec, line 43: this can be cleaned up

src/managers/CategoryManager/categoryManagerStore.js
```diff
@@ -0,0 +1,273 @@
...
+	[Actions.CATEGORY_ADD]: 'Add category',
+};
+export const useCategoryManagerStore = defineComponentStore(
+	'categoryManagerStore',
+	(props) => {
+		const {apiUrl} = useUrl('categories');
+		const {t} = useLocalize();
+		const extender = useExtender();
+		const categoryManagerConfig = extender.addFns(useCategoryManagerConfig());
+		const {data, fetch: fetchCategories} = useFetch(apiUrl, {method: 'GET'});
```
jardakotesovec, line 31: Its bit nicer when its named, so its easier to understand what it is: `{data: categories}`

src/managers/CategoryManager/categoryManagerStore.js
taslangraham: updated to expose individual action methods from store

src/managers/CategoryManager/CategoryCellToggleSubCategories.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+	<TableCell>
+		<div
+			class="flex justify-center"
+			:style="{'padding-inline-end': `${depth + 0.5}rem`}"
+		>
+			<Icon
```
jardakotesovec, line 7: I think this would semantically make sense to wrap in button. so it also gets focus when tabbing.

Also for screen reader we need some text there and hide the icon.

Similar example I found in code base, its combination of aria-hidden to hide icon that does not have meaning for screen reader user and providing text which is hidden for other users with sr-only
```
							<button
								v-if="noActions"
								class="absolute right-3 top-3 cursor-pointer bg-transparent"
								@click="onClose"
							>
								<Icon
									class="h-6 w-6 text-negative"
									icon="Cancel"
									:aria-hidden="true"
								/>
								<span class="sr-only">
									{{ t('common.close') }}
								</span>
							</button>

```

src/managers/CategoryManager/categoryManagerStore.js
jardakotesovec: Based on @touhidurabir feedback as he was working on one plugin, we agreed to always pass the 'item' for the configurations that are item specific. Even if you don't need that information, plugin might.  In this case I think it would be just getItemActions. Example with passing the galley there - https://github.com/pkp/ui-library/blob/main/src/managers/GalleyManager/galleyManagerStore.js#L40

jardakotesovec (changes requested):
Added last two things that could be improved. Than its ready for merge.

taslangraham:
> Added last two things that could be improved. Than its ready for merge.

Thanks for the review @jardakotesovec. I've made those updates.

jardakotesovec (approved):
Looking great! Appreciate quick and accurate adjustments :-).

ewhanson (approved):

PR #553, opened 2025-03-06
pkp/pkp-lib#9733 Move from moment to luxon, connect with date formats…

jardakotesovec:
… configured for journal


src/utils/convertPhpDateTimeFormatToLuxon.js
```diff
@@ -0,0 +1,160 @@
+/**
+ * PHP datetime format conversion to Luxon
+ * Generated with LLM, so probably overkill implementation,
+ * but works as expected
+ * */
+
+const phpFormatSpecifiers = new Set([
```
blesildaramirez, line 7: It seems that this is very similar to `phpToLuxonMap`. Perhaps we can just use it and change this from:
 ```
phpFormatSpecifiers.has(char)
```
to

```
Object.keys(phpToLuxonMap).includes(char)
```

blesildaramirez (approved):
Just a minor suggestion, but I could try to change it when I update the `DateTimeForm.vue`

PR #579, opened 2025-04-02
pkp/pkp-lib#10779 [stable-3_5_0] Redesigned Workflow rcX - Help modal

blesildaramirez:


src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -46,7 +46,7 @@
 								</div>
 
 								<div class="pkpDropdown__section">
-									<div class="app__userNav__loggedInAs">
+									<div class="mb-1 mt-1 pl-2 pr-2 text-base-normal leading-5">
 										{{
```
jardakotesovec, line 50: Yeah.. that was quite a shortcut to use the userNav class here :-).

src/components/Modal/SideModal.vue
```diff
@@ -1,6 +1,11 @@
 <template>
 	<TransitionRoot as="template" :show="open">
-		<HLDialog as="div" class="relative z-10" @close="handleClose">
+		<HLDialog
+			as="div"
+			class="relative z-10"
+			:initial-focus="closeModalButton"
+			@close="handleClose"
```
jardakotesovec, line 7: This is to avoid default behaviour, which would put focus on the top bar, right?

src/components/TopNavActions/TopNavActions.vue
```diff
@@ -0,0 +1,243 @@
...
+	pkp.context?.supportedLocales || {},
+).map(([key, value]) => {
+	const {pageUrl} = useUrl(`user/setLocale/${key}`);
+	return {key, value, href: pageUrl.value};
+});
+
+const currentLocale = pkp.context?.currentLocale;
+const helpUrl = pkp.context?.helpUrl;
+
+const getAnimationStyle = (buttonType) => {
```
jardakotesovec, line 175: This seems good use case for computed?  Technically the difference is that computed result is cached and recalculated only if dependencies changes (but the dependencies needs to be reactive). With function it recalculates on every render.

src/components/TopNavActions/TopNavActions.vue
```diff
@@ -0,0 +1,243 @@
...
+
+const supportedLocales = Object.entries(
+	pkp.context?.supportedLocales || {},
+).map(([key, value]) => {
+	const {pageUrl} = useUrl(`user/setLocale/${key}`);
+	return {key, value, href: pageUrl.value};
+});
+
+const currentLocale = pkp.context?.currentLocale;
+const helpUrl = pkp.context?.helpUrl;
```
jardakotesovec, line 173: This is fine, but in general trying to put all these behind composables. For this maybe useApp could be good one?

src/components/TopNavActions/TopNavActions.vue
jardakotesovec: Oh nevermind.. I see that you need to have the buttonType argument.. now I understand why you made it as function..   Thats probably fine.  To be able make it computed property it would require additional subcomponents.

src/managers/ParticipantManager/ParticipantManager.vue
```diff
@@ -87,4 +101,6 @@ const Components = {
 const participantManagerStore = useParticipantManagerStore(props);
 
 const {t} = useLocalize();
+const {isUserLoggedInAs, getCurrentUserFullName} = useCurrentUser();
+const currentuserFullName = getCurrentUserFullName();
 </script>
```
jardakotesovec, line 106: Bit nitpicking, but aiming to keep all the logic in store, rather than spread it in component&store. Can we move this to the store?

src/composables/useCurrentUser.js
```diff
@@ -16,6 +16,53 @@ export function useCurrentUser() {
...
+		return pkp.currentUser.fullName;
+	}
+
+	function getCurrentUserInitials() {
+		return pkp.currentUser.initials;
+	}
+
+	function getUserLoggedInAsUserName() {
+		if (!isUserLoggedInAs()) {
+			return '';
```
jardakotesovec, line 33: Super minor.. but would probably return `null` if its not existent.

src/managers/ParticipantManager/useParticipantManagerActions.js
```diff
@@ -136,21 +138,13 @@ export function useParticipantManagerActions() {
...
-							redirectUrl = authorRedirectUrl.value;
-						} else {
-							const {pageUrl: editorialRedirectUrl} = useUrl(
-								`dashboard/editorial?workflowSubmissionId=${submission.id}`,
-							);
-							redirectUrl = editorialRedirectUrl.value;
-						}
-						const {redirectToPage} = useUrl(
-							`login/signInAsUser/${participant.id}?redirectUrl=${encodeURIComponent(redirectUrl)}`,
+						const redirectUrl = getDashboardLoginAsUrl(
```
jardakotesovec, line 141: great to see this logic has moved to reusable composable

jardakotesovec (changes requested):
This is great work, in many ways improvement. I added just some borderline nitpicking. But did not find anything worrying :-).

src/components/ListPanel/doi/DoiListPanel.vue
blesildaramirez: Yep, the rules applied to `userNav` class translate to the used tailwind classes in the changes.

src/components/Modal/SideModal.vue
blesildaramirez: Yes. I wanted to avoid the focus on the top navbar when the side modal opens. So behavior would be the same as currently (focus is set on the close modal button)

src/composables/useCurrentUser.js
```diff
@@ -16,6 +16,53 @@ export function useCurrentUser() {
...
+		return pkp.currentUser.initials;
+	}
+
+	function getUserLoggedInAsUserName() {
+		if (!isUserLoggedInAs()) {
+			return null;
+		}
+
+		const userLoggedInAs = getUserLoggedInAs();
+		return userLoggedInAs?.username || '';
```
jonasraoni, line 37: This should cover the `null` return

```suggestion
		const userLoggedInAs = getUserLoggedInAs();
		return userLoggedInAs?.username || '';
```

Alternatively (in case the `if` is required for some reason...), I think you can assume the user is accessible


```suggestion
		if (!isUserLoggedInAs()) {
			return null;
		}

		const userLoggedInAs = getUserLoggedInAs();
		return userLoggedInAs.username;
```

src/composables/useCurrentUser.js
```diff
@@ -16,6 +16,53 @@ export function useCurrentUser() {
...
+	}
+
+	function getUserLoggedInAsInitials() {
+		if (!isUserLoggedInAs()) {
+			return null;
+		}
+
+		const userLoggedInAs = getUserLoggedInAs();
+		return userLoggedInAs?.initials || '';
```
jonasraoni, line 46: This should cover the `null` return

```suggestion
		const userLoggedInAs = getUserLoggedInAs();
		return userLoggedInAs?.initials;
```

Alternatively (is the `if` is required for some reason...)


```suggestion
		if (!isUserLoggedInAs()) {
			return null;
		}

		const userLoggedInAs = getUserLoggedInAs();
		return userLoggedInAs.initials;
```

src/composables/useUserAuth.js
```diff
@@ -0,0 +1,62 @@
...
+		const submissionId = new URLSearchParams(window.location.search).get(
+			'workflowSubmissionId',
+		);
+
+		// if the user is currently on a submission page, build the redirect url to the same submission page
+		const submissionRedirect = submissionId
+			? getDashboardLogoutAsUrl(submissionId)
+			: '';
+
+		return submissionRedirect || 'login/signOutAsUser';
```
jonasraoni, line 45: The function always return a value, so perhaps

```suggestion
		return submissionId
			? getDashboardLogoutAsUrl(submissionId)
			: 'login/signOutAsUser';
```

src/composables/useUserAuth.js
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	function getLogoutUrl() {
+		const {isUserLoggedInAs} = useCurrentUser();
+		if (isUserLoggedInAs()) {
+			return getLogoutAsUrl();
+		}
+
+		return 'login/signOut';
```
jonasraoni, line 54: Just to follow the pattern


```suggestion
		const {isUserLoggedInAs} = useCurrentUser();
		return isUserLoggedInAs() ? getLogoutAsUrl() : 'login/signOut';
```

jonasraoni (approved):
Just skimmed through the code and left non-important comments.

PR #585, opened 2025-04-10
pkp/pkp-lib#11181 remove external link from ror autosuggest field

bozana:
s. https://github.com/pkp/pkp-lib/issues/11181


bozana:
submodule update will come with my next PR...

PR #591, opened 2025-04-14
pkp/pkp-lib#11275 (stable-3_5_0) [Create New Submission] | Submission workflow - Submission title, Resolve TinyMCE wrapping text in <div> that breaks formatting 

blesildaramirez:


src/components/Form/fields/FieldRichText.vue
```diff
@@ -98,7 +98,7 @@ export default {
 							// to use this as the wrapper, and then remove as a workaround to a
 							// deprecated option & value "{ forced_root_block: '' }"
 							// @see https://github.com/tinymce/tinymce/discussions/7342
-							ed.content = ed.content.replace(/^<div>(.*)<\/div>$/, '$1');
+							ed.content = ed.content.replace(/<\/?div>/g, '');
```
touhidurabir, line 101: This works perfectly but if we consider some unusual case such as 
```
<div>I am a string <DIV class="test"><b>Bold test</b></DIV> and some more <i>more italic</i></div >
```
basically case sensitive and extra attributes, then we can do it like
```js
content.replace(/<\/?div[^>]*>/gi, '')
```

another approach that I used previously(but much shower) is 
```js
const parser = new DOMParser();
const doc = parser.parseFromString(content, 'text/html');
const divs = doc.querySelectorAll('div');
divs.forEach(div => div.replaceWith(...div.childNodes));
return doc.body.innerHTML;
```
just some consideration but I don't think tinymce add something such unusual . Up to you if you want to adjust .

touhidurabir (approved):

src/components/Form/fields/FieldRichText.vue
blesildaramirez: Thank you for checking this!
Yeah, we define the value for `forced_root_block` and it would be unusual if TinyMCE sets it to uppercase <DIV>. So I will go ahead and merge.

PR #601, opened 2025-04-19
pkp/pkp-lib#11290 Create DialogBody component to extend custom UI for dialogs

blesildaramirez:


src/components/ListPanel/doi/DoiFailedActionDialogBody.vue
```diff
@@ -1,14 +1,18 @@
 <template>
-	<p>{{ t('manager.dois.update.partialFailure') }}</p>
-	<ul class="list-inside list-disc">
-		<li v-for="(errorMessage, index) in failedDoiActions" :key="index">
-			{{ errorMessage }}
-		</li>
-	</ul>
+	<PkpDialogBody>
```
jardakotesovec, line 2: Not big deal - but using prefix only if there is conflict with native elements, like table or button. In this case I think its fine just to use <DialogBody>

src/components/Modal/DialogBody.vue
```diff
@@ -0,0 +1,67 @@
...
+	if (typeof callback === 'function') {
+		callback(() => {
+			props.onClose?.();
+		});
+	}
+}
+
+onMounted(() => {
+	if (slots.actions) {
+		emit('has-actions', true);
```
jardakotesovec, line 64: When creating such custom dialog - maybe its better to introduce option for dialog so its flexible and we can avoid this complicated detection: `showCloseButton`.  And if thats not defined, it can fallback to current behaviour, where its checking for the `actions` prop.  And if its explicitely set than it would follow this setting.

src/components/Modal/Dialog.mdx
```diff
@@ -28,6 +28,33 @@ The style of the modal can be changed by passing the prop `modalStyle`. It accep
...
+This wrapper provides an `#actions` slot where you can define more complex logic, such as:
+
+- Validation before closing the modal
+- Showing a loading spinner
+- Conditional button states
+
+```vue
+<PkpDialogBody>
+	<p>Are you sure you want to delete?</p>
+	<template #actions>
```
jardakotesovec, line 44: The idea with actions slot is that by default it gives you positioning for buttons consistent with Dialog, but you can put there what you want and hook up to these buttons for any logic.

So I would still recommend to use the default slot for any additional content, with the exception of the buttons.

jardakotesovec (changes requested):
Looking good, suggested to tweaks.

src/components/Modal/Dialog.mdx
blesildaramirez: Thank you, I've addressed this and also removed Spinner logic to modalStore.

src/components/Modal/DialogBody.vue
blesildaramirez: I added the `showCloseButton` as prop, so complex components that use action slots and want to remove the x close button can opt to set it to false, so it will not be shown.

jardakotesovec (approved):
Yay!

PR #608, opened 2025-04-29
pkp/pkp-lib#11339 add current URL as source parameter for language ch…

bozana:
s. https://github.com/pkp/pkp-lib/issues/11339

### DIFF
--- a/src/components/TopNavActions/TopNavActions.vue
+++ b/src/components/TopNavActions/TopNavActions.vue
@@ -53,7 +53,7 @@
 					class="mb-2 min-w-52 max-w-[20em]"
 				>
 					<div
-						v-if="supportedLocales.length > 1"
+						v-if="getSupportedLocalesList().length > 1"
 						class="-ml-2 -mr-2 border-b border-b-light pb-2 pl-2 pr-2"
 					>
 						<div class="ps-2 text-base-bold leading-6 text-secondary">
@@ -61,7 +61,7 @@
 						</div>
 						<ul>
 							<li
-								v-for="locale in supportedLocales"
+								v-for="locale in getSupportedLocalesList()"
 								:key="locale.key"
 								:class="actionLinkStyle"
 							>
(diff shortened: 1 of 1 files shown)


bozana:
Hi @jardakotesovec and @blesildaramirez, could (one of) you take a look at this change?
Thanks a lot!

src/components/TopNavActions/TopNavActions.vue
```diff
@@ -168,7 +168,7 @@ const {getSupportedLocales, getCurrentLocale, getHelpUrl} = useApp();
 const supportedLocales = Object.entries(getSupportedLocales() || {}).map(
 	([key, value]) => {
 		const {pageUrl} = useUrl(`user/setLocale/${key}`);
-		return {key, value, href: pageUrl.value};
+		return {key, value, href: pageUrl.value + '?source=' + document.URL};
```
bozana, line 171: Hi @jardakotesovec, how come I always get `...dashboard/editorial?currentViewId=assigned-to-me` as document.URL, even if I am on `...dashboard/editorial?currentViewId=needs-editor`, do you know?

bozana: Hmmm... I see... Only if a page reload happens in the browser, then those query and anchor parts of the URL are correct here... But when I am on a one nav item or one tab, then on another, only the first one is always returned in document.URL...
Hmmm... I suppose we cannot change anything about it, or? :thinking:

jardakotesovec: Thats because with this implementation it gets calculated just once when the TopNav is rendered. 

Its bit tricky as its not straighforward to track url changes reliably to make it nicely reactive.

So I suggest following approach.. just turning it to the function..  so it gets called when the dropdown gets opened.

```js
function getSupportedLocalesList() {
	return Object.entries(getSupportedLocales() || {}).map(([key, value]) => {
		const {pageUrl} = useUrl(`user/setLocale/${key}`);
		return {key, value, href: pageUrl.value + '?source=' + document.URL};
	});
}
```

Which can be used as
```html
					<div
						v-if="getSupportedLocalesList().length > 1"
						class="-ml-2 -mr-2 border-b border-b-light pb-2 pl-2 pr-2"
					>
						<div class="ps-2 text-base-bold leading-6 text-secondary">
							Change Language
						</div>
						<ul>
							<li
								v-for="locale in getSupportedLocalesList()"
								:key="locale.key"

```

asmecher (approved):

src/components/TopNavActions/TopNavActions.vue
bozana: Hi @jardakotesovec, cooooool ! :sunglasses: :smiley:

bozana:
If no objection, @jardakotesovec, I would merge these changes when the tests pass...

jardakotesovec (approved):

jardakotesovec:
@bozana no objections :-)

PR #622, opened 2025-05-19
pkp/pkp-lib#9890 Improve SideMenu API for less confusion

jardakotesovec:


blesildaramirez (approved):
Looks good on my end. I also tested locally and the dashboard and workflow pages work.

PR #629, opened 2025-05-27
data citations field authors

GaziYucel:


src/components/Form/fields/FieldAuthors.vue
```diff
@@ -0,0 +1,187 @@
...
+		type: String,
+		default: null,
+	},
+	/** Current value of the field */
+	value: {
+		type: Array,
+		default: () => [],
+	},
+	/** Default locale of the form */
+	primaryLocale: {
```
jardakotesovec, line 116: I think there is also opportunity to clean up some of the props - which these field does not care about. 
I expect these to be - primaryLocale, locales,

src/components/Form/fields/FieldAuthors.vue
```diff
@@ -0,0 +1,187 @@
...
+function updateRow(rowIndex, fieldName, fieldValue) {
+	currentValue.value = currentValue.value.map((row, index) => {
+		if (index !== rowIndex) {
+			return row;
+		}
+		let updatedRow = {
+			...row,
+			[fieldName]: fieldValue,
+		};
+		updatedRow.displayName = (
```
jardakotesovec, line 179: Why the displayName was introduced? Can't see where thats used.

src/components/Form/fields/FieldAuthors.vue
```diff
@@ -0,0 +1,187 @@
...
+					</TableCell>
+					<TableCell>
+						<Button :is-link="true" @click="deleteRow(index)">
+							{{ t('common.delete', []) }}
+						</Button>
+					</TableCell>
+				</TableRow>
+			</TableBody>
+			<template #bottom-controls>
+				<Button @click="addRow()">
```
jardakotesovec, line 67: Nothing wrong with using Button like this - but can I ask you to rename it to PkpButton? We do that for couple of exceptions where the name is same as existing html element - like button, table, form. So its more obvious that its our component, and not native html element. Thanks!

jardakotesovec (changes requested):
Just couple more minor things. Thanks!

PR #630, opened 2025-05-27
pkp/pkp-lib#11398 Use fullscreen loaded to prevent clicks before the …

jardakotesovec:
…email template is fetched


blesildaramirez (approved):
Tested locally and the fix works well.

PR #640, opened 2025-06-11
pkp/pkp-lib#11484 Allow html in tooltip, but still sanitize it

jardakotesovec:


src/components/Tooltip/Tooltip.vue
```diff
@@ -19,6 +20,12 @@
 import 'floating-vue/dist/style.css';
 import Icon from '@/components/Icon/Icon.vue';
 
+import DOMPurify from 'dompurify';
```
blesildaramirez, line 23: I think we can just import the directive here?
```suggestion
import {sanitizeHtml} from '@/directives/stripUnsafeHtml';
```

src/components/Tooltip/Tooltip.vue
```diff
@@ -48,6 +55,9 @@ export default {
 		},
 	},
 	computed: {
+		tooltipContent() {
+			return DOMPurify.sanitize(this.tooltip, sanitizeConfig);
```
blesildaramirez, line 59: And here we can call the sanitizeHtml from the directive:
```suggestion
			return sanitizeHtml(this.tooltip);
```

then we can delete the `sanitizeConfig` variable.

blesildaramirez (changes requested):
Tested this and works, but would I would suggest to reuse the `stripUnsafeHtml` directive on the Tooltip component.

PR #643, opened 2025-06-11
pkp/pkp-lib#11437 reviewers are always displayed on the masthead

bozana:
s. https://github.com/pkp/pkp-lib/issues/11437


src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -163,16 +162,34 @@ import {useFetch} from '@/composables/useFetch';
...
+
+/**
+ * Role options
+ */
+function getRoleOptions() {
+	var returnArray = [];
+	Object.values(props.userGroups).forEach((element, i) => {
+		returnArray[i] = {
+			value: element.userGroupId,
+			label: 'bozana' + element.userGroupId, //localize(element.name)
```
taslangraham, line 187: accidental test value, `'bozana'` left here

taslangraham, line 187: I'm not entirely sure, but maybe the `nameLocaleKey` property could be used here. Another option could be to provide the component with a `defaultLocale` prop and use that to access the correct localized name on the group. Maybe @jardakotesovec can provide some input here.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -163,16 +162,34 @@ import {useFetch} from '@/composables/useFetch';
...
 const allUserGroupsToAdd = computed(
 	() => store.invitationPayload.userGroupsToAdd,
 );
-updateWithSelectedUserGroups(props.userGroups);
+updateWithSelectedUserGroups(roleOptions);
+
+/**
+ * Role options
+ */
+function getRoleOptions() {
```
taslangraham, line 182: This method can be converted into a [computed property](https://vuejs.org/guide/essentials/computed), removing the need to have the `roleOptions` const.

src/pages/userInvitation/UserInvitationUserGroupsTable.vue
```diff
@@ -163,16 +162,34 @@ import {useFetch} from '@/composables/useFetch';
 
 const props = defineProps({
 	userGroups: {type: Object, required: true},
+	reviewerUserGroupIds: {type: Array, required: true},
 });
 
+const roleOptions = getRoleOptions();
+const reviewerRoleIds = props.reviewerUserGroupIds;
```
taslangraham, line 169: You can access the `reviewerUserGroupIds` prop directly without needing to store in a second, `reviewerRoleIds`, variable.

example:
**In template** 
```
<TableCell>				
{{
    reviewerUserGroupIds.includes(currentUserGroup.id) .....
}}
</TableCell>
``` 
**in JS** - `props.reviewerUserGroupIds`

PR #652, opened 2025-06-23
pkp/pkp-lib#11291 [Implementation] Discussions & Tasks

blesildaramirez:


src/components/Table/TableCell.vue
```diff
@@ -50,6 +51,31 @@ const classes = computed(() => {
...
+
+const tableContext = inject('tableContext');
+const registerCell = inject('registerCell');
+const columnIndex = ref(0);
+const currentColGroupId = ref('');
+
+onMounted(() => {
+	if (registerCell) {
+		columnIndex.value = registerCell();
+	}
```
jardakotesovec, line 66: You are relying just on the order how things are rendered, right? That would probably work most of the time.
But what about when new data comes in - and changes are applied to the table. Than some cells/rows can get mounted and unmounted. And I don't think that relying on the currentColGroupId will be reliable at that moment.

I can't think of other way than to keep relationship between colGroup and the Cells than have it as parent/child relationship to be able to use provide/inject. Otherwise it would require to pass something to them explicitely via props as I don't think there is reliable mechanism to tell under which colGroup I am - being just sibling.

PR #653, opened 2025-06-26
pkp/pkp-lib#11291 Add No Items row when using TableRowGroup and no items for that group

blesildaramirez:


src/components/Table/TableRowGroup.vue
```diff
@@ -1,13 +1,38 @@
 <template>
 	<slot />
+	<tr v-if="isGroupEmpty">
+		<td
+			:colspan="tableContext.columnsCount.value"
+			:headers="groupId"
+			class="border-x border-b border-light p-5 text-base-normal"
```
jardakotesovec, line 7: No strong opinion.. was just wondering whether using TableCell here which has the styling would be good fit.. but not sure..

blesildaramirez: Updated this to add a prop to customize the TableCell padding: https://github.com/pkp/ui-library/pull/653/commits/543d04d32db11e08e598be9fd7dab847bd901932

PR #656, opened 2025-07-07
pkp/pkp-lib#11577 [stable-3_5_0] [UI] Top menu items are only clickable on text

blesildaramirez:


kaitlinnewson (approved):
Hi @blesildaramirez, this looks good to me and fixes the issue in my testing.

I know it's out of the scope of this issue, but I see there is a label for "Change Language" here that is not localized so is always displayed in English.

blesildaramirez:
Hi @kaitlinnewson, update here: Alec suggested to have the localization fixed on 3.5, but we should do that after releasing `3.5.0-1`. So I'm gonna have another PR to fix that for 3.5 once we're able to release the 3.5.0-1 patch.

PR #659, opened 2025-07-09
pkp/pkp-lib#11327 Implement public facing UI for user comments

taslangraham:
For pkp/pkp-lib#11327


src/components/UserComment/UserComment.vue
```diff
@@ -0,0 +1,270 @@
...
+	pageCount.value = pagination.value.pageCount;
+}
+
+// Fetch comments after a short delay to ensure `pkp` global var is loaded before use
+onMounted(() => {
+	setTimeout(async () => {
+		currentUser.value = pkp.currentUser;
+		await fetchComments();
+		isLoading.value = false;
+	}, 500);
```
taslangraham, line 266: When the component is mounted the `pkp` var is not immediately available. As such, when using any composable that references it (e.g: `useUrl`, `useFetchPaginated`) an error is thrown. An example would be this [error](https://github.com/pkp/ui-library/blob/54b86d7b2c100a85e74c005c9fcc264122ea2f77/src/composables/useUrl.js#L11-L13) from the useUrl composable.

As a workaround, I've added a timeout  before proceeding to use these composables

src/components/UserComment/UserComment.vue
```diff
@@ -0,0 +1,270 @@
...
+	currentReportForm.value = getCommentReportForm(
+		comments.value.find((comment) => comment.id === commentId),
+	);
+
+	// The report submission logic below will not be needed once pkpForm component is used
+	const {apiUrl} = useUrl(`comments/${commentId}/reports`);
+
+	const {fetch: postReport, isSuccess} = useFetch(apiUrl, {
+		method: 'POST',
+		body: {note: 'Inappropriate content'},
```
taslangraham, line 235: Added here for testing purpose.. Will be removed once `pkpform` is being used

src/frontend/components/PkpUserComment/PkpUserCommentsList.vue
```diff
@@ -0,0 +1,479 @@
...
+			v-if="!isLatestPublication"
+			class="pkpUserCommentsList__commentNotAllowed"
+		>
+			<PkpIcon class="pkpUserCommentsList__errorIcon" icon="Error" />
+			<span>{{ t('userComment.discussionClosed') }}</span>
+		</div>
+
+		<div v-if="isLatestPublication && !!currentUser">
+			<PkpTextarea
+				placeholder="What do you think about this publication? Type your comments here."
```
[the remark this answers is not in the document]
jardakotesovec: Yes.. and also label is essential for accessibility.
So we need to pass the same text as label. Because screen readers are somewhat inconsistent how they present placeholder..   And add prop on PkpTextarea to hide it (only for screenreaders).

src/frontend/components/PkpUserComment/PkpUserCommentsList.vue
```diff
@@ -0,0 +1,479 @@
...
+import PkpButton from '@/frontend/components/PkpButton/PkpButton.vue';
+import PkpIcon from '@/frontend/components/PkpIcon/PkpIcon.vue';
+import PkpTextarea from '@/frontend/components/PkpTextarea/PkpTextarea.vue';
+import PkpDropdownActions from '@/frontend/components/PkpDropdownActions/PkpDropdownActions.vue';
+import PkpUserCommentReportModal from '@/frontend/components/PkpUserComment/PkpUserCommentReportModal.vue';
+import PkpAvatarInitials from '@/frontend/components/PkpAvatarInitials/pkpAvatarInitials.vue';
+import {usePkpModal} from '@/frontend/composables/usePkpModal';
+import {usePkpFetch} from '@/frontend/composables/usePkpFetch';
+import {useUrl} from '@/frontend/composables/usePkpUrl';
+import {useDate} from '@/composables/useDate';
```
jardakotesovec, line 129: I was hesitant to bring useDate as its quite sizeable..  But it would be good to have it..  so lets stick with it..  lets just make copy as well.

And I will iterate on it later to see whether I can make the dependency smaller.. or fine other way to do it.. but its good to have this abstracted in composable.

src/frontend/components/PkpUserComment/PkpUserCommentsList.vue
```diff
@@ -0,0 +1,479 @@
...
+
+/**
+ * Action to be performed after a comment is saved.
+ * @param data - The saved comment data.
+ */
+function onCommentSaved(data) {
+	commentText.value = '';
+	// Add the new comment to the top of the comments list.
+	comments.value.unshift(data);
+}
```
jardakotesovec, line 344: Once we update something via API - its best to reload the data and consider that as source of the truth.

Doing optimistic update on top of it is fine to make it bit faster.  But in general for simplicity - I basically just call API for update and than refresh the data. So here doing the same would be fine as well.  As long as there is indication that the submit is being processed its fine to show it with bit of delay.

src/frontend/components/PkpUserComment/PkpUserCommentsList.vue
```diff
@@ -0,0 +1,479 @@
...
+		pageSize: props.itemsPerPage,
+	});
+
+	await fetch();
+
+	comments.value = [...comments.value, ...items.value];
+	pageCount.value = pagination.value.pageCount;
+	showMoreCommentsCount.value =
+		showMoreCommentsCount.value - props.itemsPerPage;
+}
```
jardakotesovec, line 367: Looks good..  useFetchPaginated was not really created for "show more"..  its done for using pagination..    so there would be opportunity to abstract that better..   but I think this is fine.

jardakotesovec (changes requested):
To give you heads up. After we merge this version.. I will start gradually refining some of the components there.. and possibly reorganise things once I have good styling strategy settled. I would probably make these adjustments myself to see it in practice and just ask you for review. Lets see how that goes.

src/frontend/components/PkpUserComment/PkpUserCommentsList.vue
taslangraham: I agree, the best simpler approach would be to refresh the data via API call.

However, I ran into a few concerns/roadblocks trying to refresh an entire list built from multiple paginated responses.

The challenge: when a user adds a new comment, it will be available on page 1 of the paginated results (most recent items comes first). To update the list via API to include the new comment, while preserving items already shown, I’d need to refetch items from page 1 up to the current page.

**Options I’ve considered:**

1. Refetch pages 1 to current
    - would work, but not performant. Also, the new item on page 1 pushes another off and we might lose an item after finish fetching because it was pushed to another page not in the range(pages 1 to current) fetched.

2. Just fetch for page 1 where the new comment is expected to be and update that portion of the list
    - there's possibility that moderators approve new comments during the same time span, and those could potentially end up on that list returned from API depending on when they were created, and potentially displacing the expected new comment

3. Single request with `pageSize = current length of comments + 1` (+1 to compensate for the newly added comment)
    - More efficient, but there's possibility that  moderators approve new comments during the same time span, and those could potentially end up on that list returned from API depending on when they were created, thus potentially pushing more items off the final updated list shown to user.

4. Single request with `pageSize` = current length + 1 full page size
    - Benefits of number 3, but reduces the potential of a newly approved comment displacing one that was already visible to the user -- this is still a possibility though. 


I initially went for the optimistic update, appending the created comment returned from the API,  due to the challenges outlined above along with the fact that a newly created comment is only visible to the creator, and we just want to display it to them to confirm creation and to inform them that it will need approval before being visible to others.

I'm probably thinking about this incorrectly, so I'm open to feedback on how to approach this

jardakotesovec: I see..  did not think about the "show more" impact here. In general "show more" is better to implemented with "cursor based" pagination.. as that ensures that newly added message (from anyone) does not mess up the list. But not sure how much effort that would be..   as we don't expect significant use - it will be fine either way.

Regarding what to do after posting new message. I think just going back to refetching "first page" and throwing rest should be fine..   and user can use show more again if they want to go back to the older messages. Since new messages is being added on top - I don't think that would be problematic.

src/frontend/components/PkpButton/PkpButton.vue
```diff
@@ -13,12 +13,15 @@ const props = defineProps({
 		type: Boolean,
 		default: false,
 	},
+	/** Use when this button represents an action such as delete, go back, revert or cancel. */
+	isWarnable: Boolean,
```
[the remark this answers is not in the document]
taslangraham: updated

PR #664, opened 2025-07-16
pkp/pkp-lib#11576 Add backoffice UI for comment moderation

taslangraham:
For pkp/pkp-lib#11576


ewhanson (changes requested):
Hey @taslangraham, looking pretty good to me. Just a few comments and one slightly larger question. 

I'm not sure if we have an established pattern for how to handle this, but the layout shift from the spinner when loading new tabs doesn't feel great. I know I'm working locally so the data loads very quickly, but I wonder if there's a better way to handle this. The loading spinner spinning while it says there are no items and then the items appearing may cause some confusion as well.

https://github.com/user-attachments/assets/c6251b88-d557-4011-9c67-14e34ad9421a

taslangraham:
> Hey @taslangraham, looking pretty good to me. Just a few comments and one slightly larger question.
> 
> I'm not sure if we have an established pattern for how to handle this, but the layout shift from the spinner when loading new tabs doesn't feel great. I know I'm working locally so the data loads very quickly, but I wonder if there's a better way to handle this. The loading spinner spinning while it says there are no items and then the items appearing may cause some confusion as well.
> 
>  Screen.Recording.2025-08-24.at.09.54.09.mov


This pattern of controlling the text shown in the table when loading data should suffice
https://github.com/pkp/ui-library/blob/9784c61c34db37fa379873e822248331b9c9df8a/src/pages/dashboard/components/DashboardTable/DashboardTable.vue#L23-L29

taslangraham:
> Hey @taslangraham, looking pretty good to me. Just a few comments and one slightly larger question.
> 
> I'm not sure if we have an established pattern for how to handle this, but the layout shift from the spinner when loading new tabs doesn't feel great. I know I'm working locally so the data loads very quickly, but I wonder if there's a better way to handle this. The loading spinner spinning while it says there are no items and then the items appearing may cause some confusion as well.
> 
>  Screen.Recording.2025-08-24.at.09.54.09.mov

The issues mentioned above and shown in the video have been fixed.

src/pages/userComments/UserCommentDetailModal.vue
```diff
@@ -0,0 +1,143 @@
...
+		<template #description>
+			{{ comment.userName }}
+		</template>
+
+		<div class="min-w-fit p-4">
+			<div class="w-full bg-secondary">
+				<div class="border-y border-e border-s border-light">
+					<div class="flex h-full">
+						<div class="flex-grow border-e border-light p-5">
+							<Panel>
```
jardakotesovec, line 22: I was thinking whether to evolve Panel or some new component to handle these sections with borders. At that time I was working on the Review details.. and you can see there are 3 different variants of that.. with different semantic structure. So in that context I gave up on that - and just went with using tailwindcss where its pretty easy to do the styling like this..  and it allows for Devika be more flexible with these designs for different kind of data. So even here I would suggest just to use tailwindcss to make it look like the design and correct semantic elements (correct heading level, p for the text etc..)





Screenshot 2025-08-26 at 11 35 48

src/pages/userComments/userCommentStore.js
```diff
@@ -0,0 +1,486 @@
...
+								`${apiUrl.value}/${report.userCommentId}/reports/${report.id}`,
+								{
+									method: 'DELETE',
+								},
+							);
+
+							await deleteReport();
+
+							if (isSuccess.value) {
+								pkp.eventBus.$emit(
```
jardakotesovec, line 163: Looks good.. I will see if I can get feedback from Devika on this. Because we basically have not managed to establish pattern about these kind of notifications..  so I will get back to you on this one tomorrow.  Nothing wrong with it - but if Devika confirms that we want to stick with this pattern - I would just make sure we have composable to trigger these.   As our composables are pretty much the API to all the interactions with the UI and API.

jardakotesovec (changes requested):
This is excellent work.

I am not sure how much demand would be for extending this UI. But it should be relatively low effort to use some of our patterns for extensibility. So they can add buttons/columns. And in fact you already have lots of these encapsulated in data structures like `commentsTableColumns` or `getReportItemActions` etc..

Pattern to follow would be that the pieces of the UI that we would want to make configurable (typically places where we can iterate over columns or components to display) you can please these 'configuration functions' to separate composable - userCommentsStoreConfig.js

You basically already used this pattern for the CategoryManager.

If its ok, I will leave it up to you which part you would make configurable, but feel free to reach out if you want more guidance on that.

General idea is that if you let them add custom action(s), custom columns and custom component to the listings..   The extensibility opportunities are great.

src/pages/userComments/userCommentStore.js
```diff
@@ -0,0 +1,486 @@
...
+				if (comment.isApproved) {
+					status.push(t('manager.userComment.approved'));
+				} else {
+					status.push(t('manager.userComment.hiddenOrNeedsApproval'));
+				}
+				if (comment.isReported) {
+					status.push(t('manager.userComment.reported'));
+				}
+
+				status = status.join(', ');
```
jardakotesovec, line 453: we even have localisation for the comma.. as apparantely some languages use different character - 

```
msgid "common.commaListSeparator"
msgstr ", "

msgid "common.semicolonListSeparator"
msgstr "; "

```

src/pages/userComments/userCommentStore.js
```diff
@@ -0,0 +1,486 @@
...
+			// Don't proceed if the current comment is already in the desired approval state.
+			if (currentComment.value.isApproved === isApproved) {
+				return;
+			}
+
+			const form = new FormData();
+			form.set('approved', isApproved);
+
+			const {isSuccess, fetch} = useFetch(
+				`${apiUrl.value}/${currentComment.value.id}/setApproval`,
```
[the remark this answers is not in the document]
taslangraham: This in fact can be done as a JSON. Thanks for pointing that out.

taslangraham:
@jardakotesovec Thanks for the review. I've made updates based on your feedback. Most importantly:
 - introduce 'config' composable to allow extensibility
 - Replace panel usage with tailwind styling

I've also introduce individual cell components for each table cell

jardakotesovec (approved):
Thanks for the extensibility improvements! Looks great.

PR #670, opened 2025-07-18
pkp/pkp-lib#11558 resolve accessibility error for charts in storybook

kaitlinnewson:


blesildaramirez (approved):
Looks good to me. We can also clean up the usages by removing the `aria-hidden="true"` attribute from the ojs/omp/ops, since it's now handled by the component.

PR #680, opened 2025-07-31
pkp/pkp-lib#11291 [Implementation] Discussions & Tasks - EDIT & VIEW

blesildaramirez:


jardakotesovec (changes requested):
Looking great - just please check where the role=group should be used.

Great progress on FormDisplay, thanks for that!

PR #683, opened 2025-08-04
pkp/pkp-lib#9295 continuous publication implementation update

touhidurabir:
for pkp/pkp-lib#9295 , second batch of PRs to update the implementation to 
1. remove the previously introduced `published` column in `publications` table
2. introduce on the `status` column of `publications` table by introducing a new status `PKPSubmission::STATUS_READY_TO_PUBLISH`
3. update the publishing/scheduling workflow by combining multiple modal into one .

### DIFF
--- a/src/components/Form/fields/FieldUpload.vue
+++ b/src/components/Form/fields/FieldUpload.vue
@@ -258,14 +258,12 @@ export default {
 			 * Add attributes to the hidden file input field so that labels and
 			 * descriptions can be accessed by those using assistive devices.
 			 */
-			if (this.$refs.dropzone) {
-				this.$refs.dropzone.dropzone.hiddenFileInput.id =
-					this.dropzoneHiddenFileId;
-				this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
-					'aria-describedby',
-					this.describedByIds,
-				);
-			}
+			this.$refs.dropzone.dropzone.hiddenFileInput.id =
+				this.dropzoneHiddenFileId;
+			this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
+				'aria-describedby',
+				this.describedByIds,
+			);
 
 			/**
 			 * Set the initial data, which can't be set in the data() function because it relies on
--- a/src/pages/workflow/components/publication/WorkflowPublicationForm.vue
+++ b/src/pages/workflow/components/publication/WorkflowPublicationForm.vue
@@ -73,6 +73,9 @@ async function metadataDataChange() {
 	await fetchForm();
 }
 
+// This handle the case where the issue selection done vai `Schedule For Publication` modal while at issue page
+// but not finalise the publication status(publishing or scheduling) at final step but close the modal
+// so we need to fetch the form again to update the issue selection field to reflect the changes.
 async function issueDataChange() {
 	await fetchForm();
 }
@@ -81,10 +84,13 @@ watch(
 	form,
 	async (newForm) => {
 		if (newForm && props.formName === 'issue' && isOJS()) {
-			const {initialize} = useWorkflowPublicationFormIssue(newForm);
+			const {initialize} = useWorkflowPublicationFormIssue(
+				newForm,
+				props.publicationpublication,
+			);
 			await initialize();
 		}
 	},
-	{immediate: true},
+	{immediate: false},
 );
 </script>
--- a/src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
+++ b/src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
@@ -2,12 +2,10 @@ import {useFetch} from '@/composables/useFetch';
 import {useUrl} from '@/composables/useUrl';
 import {computed, watch} from 'vue';
 import {useLocalize} from '@/composables/useLocalize';
-import {useWorkflowStore} from '@/pages/workflow/workflowStore';
 import {useForm} from '@/composables/useForm';
 
-export function useWorkflowPublicationFormIssue(form) {
+export function useWorkflowPublicationFormIssue(form, selectedPublication) {
 	const {t} = useLocalize();
-	const store = useWorkflowStore();
 	const {
 		setHiddenValue,
 		getHiddenValue,
@@ -21,7 +19,7 @@ export function useWorkflowPublicationFormIssue(form) {
 	const {fetch: fetchAssignments, data: assignments} = useFetch(assignmentsUrl);
 
 	const {apiUrl: assignmentStatusUrl} = useUrl(
-		`submissions/${store.submission.id}/publications/${store.selectedPublication.id}/issueAssignmentStatus`,
+		`submissions/${selectedPublication.submissionId}/publications/${selectedPublication.id}/issueAssignmentStatus`,
 	);
 	const {fetch: fetchAssignmentStatus, data: assignmentStatus} =
 		useFetch(assignmentStatusUrl);
@@ -103,14 +101,8 @@ export function useWorkflowPublicationFormIssue(form) {
 
 		const issueIdField = getField('issueId');
 		const assignmentField = getField('assignment');
-		const statusValue = getHiddenValue('status');
 
-		return (
-			issueIdField &&
-			assignmentField &&
-			statusValue !== null &&
-			statusValue !== undefined
-		);
+		return issueIdField && assignmentField;
 	}
 
 	/**
@@ -129,13 +121,13 @@ export function useWorkflowPublicationFormIssue(form) {
 			description: t('publication.assignToIssue.issueDescription'),
 			options: [],
 			size: 'large',
-			value: store.selectedPublication?.issueId,
+			value: selectedPublication?.issueId,
 			isRequired: true,
 			showWhen: ['assignment', []],
 		});
 
 		// Set initial status - will be updated by watchers when assignmentStatus data arrives
-		setHiddenValue('status', store.selectedPublication?.status);
+		setHiddenValue('status', selectedPublication?.status);
 	}
 
 	/**
@@ -161,7 +153,7 @@ export function useWorkflowPublicationFormIssue(form) {
 						![
 							pkp.const.publication.STATUS_PUBLISHED,
 							pkp.const.publication.STATUS_SCHEDULED,
-						].includes(store.selectedPublication?.status)
+						].includes(selectedPublication?.status)
 					) {
 						setHiddenValue('status', newStatus.status);
 					}
@@ -191,7 +183,7 @@ export function useWorkflowPublicationFormIssue(form) {
 					issueIdField.showWhen = ['assignment', newShowWhenIds];
 				}
 			},
-			{immediate: true},
+			{immediate: false},
 		);
 
 		// Watch current assignment option changes and update status
@@ -209,10 +201,12 @@ export function useWorkflowPublicationFormIssue(form) {
 						[
 							pkp.const.publication.STATUS_PUBLISHED,
 							pkp.const.publication.STATUS_SCHEDULED,
-						].includes(store.selectedPublication?.status)
+						].includes(selectedPublication?.status)
 					) {
 						setHiddenValue('status', null);
 					}
+
+					isInitialDataLoad = false;
 				}
 
 				if (newOption.isPublished !== null) {
@@ -232,12 +226,6 @@ export function useWorkflowPublicationFormIssue(form) {
 			},
 			{immediate: true},
 		);
-
-		// Mark initial data load as complete after a short delay
-		// which in return all initial watchers have had a chance to run
-		setTimeout(() => {
-			isInitialDataLoad = false;
-		}, 100);
 	}
 
 	/**
--- a/src/pages/workflow/composables/useWorkflowVersionForm.js
+++ b/src/pages/workflow/composables/useWorkflowVersionForm.js
@@ -321,7 +321,10 @@ export function useWorkflowVersionForm(
 		// it's in publish mode
 		// have issues
 		if (modeState.isPublishMode && issueCount > 0 && isOJS()) {
-			const {initialize} = useWorkflowPublicationFormIssue(form);
+			const {initialize} = useWorkflowPublicationFormIssue(
+				form,
+				store.selectedPublication,
+			);
 			await initialize();
 		}
 
--- a/src/utils/defineComponentStore.js
+++ b/src/utils/defineComponentStore.js
@@ -56,10 +56,6 @@ export function defineComponentStore(
 			storesMap[storeName].mountedCount = storesMap[storeName].mountedCount + 1;
 		});
 		onBeforeUnmount(() => {
-			if (!storesMap[storeName]) {
-				return;
-			}
-
 			storesMap[storeName].mountedCount = storesMap[storeName].mountedCount - 1;
 
 			if (storesMap[storeName].mountedCount === 0) {
--- a/src/components/Form/fields/FieldUploadImage.vue
+++ b/src/components/Form/fields/FieldUploadImage.vue
@@ -250,14 +250,12 @@ export default {
 			 * Add attributes to the hidden file input field so that labels and
 			 * descriptions can be accessed by those using assistive devices.
 			 */
-			if (this.$refs.dropzone) {
-				this.$refs.dropzone.dropzone.hiddenFileInput.id =
-					this.dropzoneHiddenFileId;
-				this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
-					'aria-describedby',
-					this.describedByIds,
-				);
-			}
+			this.$refs.dropzone.dropzone.hiddenFileInput.id =
+				this.dropzoneHiddenFileId;
+			this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
+				'aria-describedby',
+				this.describedByIds,
+			);
 
 			/**
 			 * Set the initial data, which can't be set in the data() function because it relies on


jardakotesovec:
@asmecher I would prefer not to merge this version, and still make the adjustments to use existing form fields, and not creating new field which combines multiple fields.

Example I provided is available on these PRs..
https://github.com/pkp/ui-library/pull/688
https://github.com/pkp/ojs/pull/5049

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,275 @@
...
+		() => assignmentDataRef.value.assignmentOptions,
+	);
+	const currentAssignment = computed(
+		() => assignmentDataRef.value.currentAssignment,
+	);
+
+	/**
+	 * Check if the required fields already exist in the form
+	 */
+	async function checkFieldsExist() {
```
jardakotesovec, line 38: What about if we just add these fields in both use cases client side - and remove these fields on the server side.. just with the comment that these gets attached client side?

Do you see some issue with that? Would like to simplify it.. as there is risk if someone is touching this in couple of months - that they won't find both places..   its always risky doing the same thing on two places.

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,275 @@
...
+
+			// Fetch current assignment status
+			const {apiUrl: assignmentStatusUrl} = useUrl(
+				`submissions/${store.submission.id}/publications/${store.selectedPublication.id}/issueAssignmentStatus`,
+			);
+			const {fetch: fetchAssignmentStatus, data: assignmentStatusData} =
+				useFetch(assignmentStatusUrl);
+			await fetchAssignmentStatus();
+
+			const assignmentOptions =
```
jardakotesovec, line 74: When possible..  good pattern for composition api & reactive data is just to implement derived data from the fetched data as computed properties.  But thats not big deal.

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,275 @@
...
+				currentStatus,
+				optionsWhichShowsIssues,
+				issueOptions,
+			};
+
+			assignmentDataRef.value = data;
+
+			return data;
+		} catch (error) {
+			console.error('Error fetching assignment data:', error);
```
jardakotesovec, line 116: As mentioned above.. you don't really have to handle this outside of the useFetch.  And its handled inside the useFetch exactly for this reason, so we don't end up with repeating error handling.. 

In terms of setting some fallback values. Simple pattern is having fetch data and computed property reading from the fetch data - and if the fetched data are empty or unavailable - thats good place to provide some fallback value.

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,275 @@
...
+		let isUpdatingIssues = false; // Flag to prevent duplicate API calls
+
+		watch(
+			currentAssignment,
+			async (newAssignmentValue, oldAssignmentValue) => {
+				if (!newAssignmentValue || isUpdatingIssues) {
+					return;
+				}
+
+				isUpdatingIssues = true; // Set flag to prevent duplicate calls
```
jardakotesovec, line 200: useFetch by default already cancels previous call if new one comes before the last finishes.

But its necessary to use it in a way, that the `useFetch` instance is there once for particular call..

Thats why this pattern is being used:
const {data: issues, fetch: fetchIssues} = useFetch(...).

where calling multiple times fetchIssue will handle that correctly.  But if you put useFetch to some function and call that repeatedly, that creates new independent useFetch every time.

src/pages/workflow/composables/useWorkflowVersionForm.js
```diff
@@ -282,10 +316,28 @@ export function useWorkflowVersionForm(
 			}),
 		);
 
+		// Issue assignment fields only visible when
+		// for OJS
+		// it's in publish mode
+		// have issues
+		if (modeState.isPublishMode && issueCount > 0 && isOJS()) {
+			await useWorkflowPublicationFormIssue(form);
```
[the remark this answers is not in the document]
touhidurabir: The main reason to make it async/wait this composable is to accommodate the both use case where if the field are defined at PHP end's then need to determine and make sure not to create those at JS end . please correct me if I am mistaken , but  use of `onMounted` here does not work properly when it's being used in another composable as `onMounted` is a Vue lifecycle hook that only works in Vue components, not in composables called from other composables or functions . In this case when use the composable from another composable , the `await` ensures the composable is fully initialized before the parent continues. probably possible to use `nextTick` or a watch on `from` itself , but I found this simpler and easier .

src/composables/useLegacyGridUrl.test.js
```diff
@@ -32,16 +32,4 @@ describe('useLegacyGridUrl', () => {
 			'http://mock/index.php/publicknowledge/$$$call$$$/grid/users/reviewer/reviewer-grid/read-review?submissionId=13&reviewAssignmentId=19&stageId=3',
 		);
 	});
-
-	test('modals.publish.AssignToIssueHandler', () => {
-		const {url} = useLegacyGridUrl({
```
jardakotesovec, line 37: Why is this one deleted?

src/pages/workflow/composables/useWorkflowVersionForm.js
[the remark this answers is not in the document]
jardakotesovec: Use of onMounted is actually unnecessary in this particular situation. I was just checking with Blessie, why she used it and she confirmed it works without it fine as well.

Theoretically if you use await within setup function - that can block rendering until you fetch the details. If its done in onMounted it does not block rendering.  But I would avoid blocking rendering based on API call..   Its not ideal to have there the field which is not fully initialised - but we have the same problem if you change type and the list issue is being fetched again. We can't (and it would be annoying UX) block rendering there.  

Thats why I was mentioning that improvement to that eventually would be to have capability to indicate that some field is being refreshing/loading so its super clear that there is something going on.   But thats something I can consider in future iterations of our form framework..  Not much for you to worry about there.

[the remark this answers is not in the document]
jardakotesovec: Therefore I think you can think about adding the fields and fetching data as two independent operations, so the field gets eventually its data after the API call is done.

src/composables/useLegacyGridUrl.test.js
touhidurabir: The assign to issue handler has been removed as we no longer have a separate controller to assign issue but part of edit publication API .

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
touhidurabir: I think that better but only one confusion . the issue form is rendered via `APP\components\forms\publication\IssueEntryForm` and then it's not adding issue fields there which may seems confusing for anyone who does not know how it's being handled . 

@asmecher do you have anything concern if the issue fields are fully constructed and handled at the client's end ?

jardakotesovec: Yeah until we moved the whole thing to client side there will be some level of confusion. Meanwhile it can be compensated with comment which would explain it. I would be fine either way - just wanted to encourage to have one implemenation.. but its not complex implemenation so its not big deal.

We used to have tricky duplication in past for publication forms, where all the data and configuration happened in PHP for the first publication, but if you have multiple versions and switch between them - it would do same logic client side and in some cases it was not trivial logic.  And that would be very difficult to notice during testing/regressions.  So compared to this - we will be fine either way :-).

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
jardakotesovec: Also we have `debouncedMs` on the useFetch in case you would need that. I had some complex scenario where it was difficult to avoid double triggering of data fetching.   And by default it just cancels the first one. But with debouncing it just makes one request.

src/composables/useLegacyGridUrl.test.js
touhidurabir: if we want to keep it , better to use a different one that exists . No problem keep this one that does not exists either .

src/components/Form/Form.vue
```diff
@@ -490,18 +493,29 @@ export default {
...
-			const newFields = this.fields.map((field) => {
-				if (field.name === name) {
-					if (localeKey) {
-						field[prop][localeKey] = value;
-					} else {
-						field[prop] = value;
+			// Check if this is a hidden field or normal field
+			const isHiddenField = this.hiddenFields && name in this.hiddenFields;
+
+			if (isHiddenField) {
```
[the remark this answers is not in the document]
touhidurabir: the use case is if based on update of a normal field , need to update the hidden field in the client side . this is something we can do with one normal field to another right now but not to the hidden field .

src/components/Form/Form.vue
```diff
@@ -13,6 +13,7 @@
 			type="hidden"
 			:name="name"
 			:value="value"
+			@change="fieldChanged"
 		/>
```
jardakotesovec, line 17: I think this one is not needed if we stick with explicit programatic changes with useFrom from outside of the form.

src/components/Form/FormGroup.vue
```diff
@@ -45,6 +45,7 @@
 								:locales="availableLocales"
 								@change="fieldChanged"
 								@set-errors="setFieldErrors"
+								@set-field-required="setFieldRequired"
```
jardakotesovec, line 48: Is this used? Can't find where this would be emitted.

public/globals.js
```diff
@@ -96,10 +96,18 @@ window.pkp = {
 		ROLE_ID_SUB_EDITOR: 17,
 		ROLE_ID_SUBSCRIPTION_MANAGER: 2097152,
 		STAGE_STATUS_SUBMISSION_UNASSIGNED: 1,
-		STATUS_QUEUED: 1,
-		STATUS_PUBLISHED: 3,
-		STATUS_DECLINED: 4,
-		STATUS_SCHEDULED: 5,
+		submission: {
+			STATUS_QUEUED: 1,
```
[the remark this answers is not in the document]
jardakotesovec: Oki.. I see some new publication statuses now that were introduced.. so I get it now..  so question is really just the naming convention..

src/pages/workflow/composables/useWorkflowActions.js
```diff
@@ -116,72 +133,10 @@ export function useWorkflowActions() {
...
-					},
-				},
-			);
-		} else {
-			workflowScheduleForPublication(
-				{submission, selectedPublication},
-				finishedCallback,
-			);
-		}
+		workflowScheduleForPublication(
```
jardakotesovec, line 136: This one is here just for omp&ops at this point right?

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,208 @@
...
+				}
+			}
+		});
+
+		// Watch issue options and apply to form
+		watch(issueOptions, (newIssueOptions) => {
+			const issueIdField = getField('issueId');
+			if (issueIdField) {
+				issueIdField.options = newIssueOptions;
+				issueIdField.isRequired = newIssueOptions.length > 0;
```
jardakotesovec, line 186: Are we sure about this one?

In Issue Assignment if they select intention of assign to issue. This would let them not do it if there is no issue. Does not cause this problem further down the line?

src/utils/defineComponentStore.js
```diff
@@ -56,6 +56,10 @@ export function defineComponentStore(
 			storesMap[storeName].mountedCount = storesMap[storeName].mountedCount + 1;
 		});
 		onBeforeUnmount(() => {
+			if (!storesMap[storeName]) {
```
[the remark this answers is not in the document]
jardakotesovec: Can you try to remove it to see whether its still needed? If this is still needed let me know.. and I will look into why.. as this is suggesting there is some bug or misuse..  So I rather figure out whats going on than just silence it ..

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	/**
+	 * Initialize the composable - fetch data, create fields, and set up watchers
+	 */
+	async function initialize() {
+		await fetchAssignments();
+		await fetchAssignmentStatus();
+		await fetchIssues();
+
```
jardakotesovec, line 198: Is there reason to have assignment status coming from its own endpoint? Could it be maybe part of the publication object? For example issue is in the publication object even though its OJS specific concept.


I will discuss with Devika some option how to handle that some fields are still initialising. I think we will have two options - to indicate progress indicator for whole form or individual fields. Thats out of the scope of this work and will address that later. But its important to acknowledge there is some loading scenario.

With this implementation - it shows the Version specific fields immediately and makes the background call for issue assignments and than shows the rest.  Lets show all of them from the beginning - even though it means that the assignment and issues field will take bit to load data and be accurate.

Benefit of declarative approach of these UI libraries is that the order does not matter much - the final state will be accurate as the data are being fetched if the computed properties are done in this way.

So basically if you remove all the awaits - it makes the fetches in parallel (that what we want) and the computed properties and watchers make sure it will eventually end up in the right state. And only missing part is the "progress indicator" which as I mentioned I address separately.

jardakotesovec (changes requested):
Hi Touhidur,
Its lot of changes, I tried to get through everything in detail..  let me know if something is unclear.

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,208 @@
...
+			return [];
+		}
+
+		if (!issues.value || !issues.value.items) {
+			return [];
+		}
+
+		return issues.value.items.map((issue) => ({
+			value: issue.id,
+			label: sanitizeHtml(issue.identification),
```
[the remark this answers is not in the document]
touhidurabir: right , I forgot that it's being used a vue component internally .

src/pages/workflow/composables/useWorkflowVersionForm.js
```diff
@@ -92,20 +96,39 @@ export function useWorkflowVersionForm(
...
 
+		// Prepare request body with version data
+		const requestBody = {
+			versionStage: formData.versionStage,
+			versionIsMinor: formData.versionIsMinor,
+		};
+
+		if (isOJS() && modeState.isPublishMode) {
+			// if the issue count is 0, e.g. issueless context,
+			// we can safely set the issueId to null and status to STATUS_READY_TO_PUBLISH
```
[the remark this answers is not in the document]
touhidurabir: No actually , if there no issue for the context(Journal), it will not be visible and thats why we have that check . But if there is issue , the default option until any selection and confirmation made is to assign to current/back issue if available and then no issue .

src/pages/workflow/components/publication/WorkflowPublicationForm.vue
```diff
@@ -55,16 +59,31 @@ const displayNoFieldsEnabled = computed(() => {
 	return false;
 });
 
+const {set, form} = useForm(publicationForm);
+
 const customFns = {
 	metadata: metadataDataChange,
+	issue: issueDataChange,
```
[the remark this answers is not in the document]
touhidurabir: This is to handle just one very special case when user in the issue page and update issue selection (assignment or issue itself in dropdown) and maybe reload the page or come back later . Then again from issue page, click on the `Schedule For Publication` , and there update the issue assignment/selection, set the version details and `confirm` . But then in the final stage instead of publish/schedule , just close the modal , then the issue assignment/selection that updated in the modal does not reflect on the issue page without reloading . So this the option I found . And this is the only case to handle by this change .

src/pages/workflow/composables/useWorkflowPublicationFormIssue.js
```diff
@@ -0,0 +1,208 @@
...
+	});
+
+	const currentAssignmentOption = computed(() => {
+		if (!assignments.value || assignments.value.length === 0) {
+			return null;
+		}
+
+		let currentValue = getField('assignment')?.value;
+
+		// If no form value yet, try to get from store (for modal context)
```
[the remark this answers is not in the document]
touhidurabir: Interestingly , I can not remember why I added this as this is not supposed to be needed . probably added at initial stage while some watcher not working as I have needed to and later forgot to remove it after fixing those .

public/globals.js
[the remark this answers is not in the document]
asmecher: @jardakotesovec, we were looking for ways to have the constant names consistent between PHP-land and JS-land, so I didn't like the idea of e.g. prefixing `STATUS_QUEUED` with `PUBLICATION_`. This felt like a closer match to the namespacing approach we've taken PHP-side.

src/utils/defineComponentStore.js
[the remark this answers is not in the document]
touhidurabir: unfortunately still needed , to reproduce it open any submission workflow(say in the side section of `All in production stage`) , navigate to `issue` tab , perhaps perform some actions, then close it and try to open another submission workflow without navigating to other tab or reloading it . I have feeling that it has something to do with following 
```js
watch(
	form,
	async (newForm) => {
		if (newForm && props.formName === 'issue' && isOJS()) {
			const {initialize} = useWorkflowPublicationFormIssue(newForm);
			await initialize();
		}
	},
	{immediate: true},
);
```
in the `WorkflowPublicationForm.vue` but could not figure it out .

src/components/Form/fields/FieldUpload.vue
```diff
@@ -258,12 +258,14 @@ export default {
...
 			 */
-			this.$refs.dropzone.dropzone.hiddenFileInput.id =
-				this.dropzoneHiddenFileId;
-			this.$refs.dropzone.dropzone.hiddenFileInput.setAttribute(
-				'aria-describedby',
-				this.describedByIds,
-			);
+			if (this.$refs.dropzone) {
+				this.$refs.dropzone.dropzone.hiddenFileInput.id =
+					this.dropzoneHiddenFileId;
```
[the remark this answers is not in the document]
touhidurabir: I was getting console error like `can not get data dropzone on null or undefined 'this.$refs.dropzone'` at workflow `issue` which turns out due a watcher running immediately which was altering dropdown's `showWhen` value . seems I dont need the watcher to run immediately but really confused why it was causing problem there when these 2 fields kind of unrelated .

src/components/Form/fields/FieldSelectIssue.vue
```diff
@@ -1,3 +1,4 @@
+<!-- TODO: need this anymore ? remove this after testing -->
```
jardakotesovec, line 1: Any thoughts on this? I don't see FieldSelectIssue used anywhere.. so I am supportive to remove it.. and just make sure that the tests are passing.

jardakotesovec (approved):
Just two clean up suggestions. I think its ready for merging.

Thanks!

PR #690, opened 2025-08-12
pkp/pkp-lib#10553 condsider only major versions in DoiListPanel

bozana:
s. https://github.com/pkp/pkp-lib/issues/10553


src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -915,17 +926,68 @@ export default {
...
+		/**
+		 * Gets pubObject's versions/publications. Considers only latest minor versions. Assumes original object is submission.
+		 *
+		 * @param {Object} item Item being mapped
+		 * @returns {Object} Modified mapped item
+		 */
 		getVersions(item) {
+			var latestMinorPublications = [];
+			item.publications?.forEach((publication) => {
+				if (publication.versionStage == null) {
```
ewhanson, line 938: I don't know if this really matters in the big picture, but I found this whole `getVersions()` method a little hard to wrap my head around. I don't know if it's the formatting or the amount of conditionals, but a few more comments stepping out what's happening may be helpful when coming back to edit/work on this in the future. Not really needed to make it *work*, though.

ewhanson (changes requested):

PR #691, opened 2025-08-15
pkp/pkp-lib#11291 Handle task/discussion status updates, save form in FormDisplay component

blesildaramirez:
Changes include: 
1. `FormDisplay`: to wrap in `<form>` if it contains some form fields, also to display "Cancel" and "Save" buttons
2. `TableCellSelect`: added props to show confirmation dialog before changing the state of the checkbox
3. `useFormChanged`: api to detect unsaved form changes
4. Discussion View: added "Close this discussion", "Start this task", "Close this task" checkboxes
5. Add Task Details: set focus of the edit mode to the Task Information section
6. Start/Close Task/Discussion: made checkboxes in the table listing clickable

### DIFF
--- a/src/composables/useFormChanged.mdx
+++ b/src/composables/useFormChanged.mdx
@@ -28,7 +28,7 @@ Enable a warning dialog if there are unsaved changes before:
 Pass the option `warnOnClose: true` and an `onCloseFn` callback (called when the user confirms closing). This will automatically attach a `beforeunload` listener for browser tabs and register the close callback for modals.
 
 ```
-const { confirmClose } = useFormChanged(
+useFormChanged(
   form,
   additionalFields,
   onCloseFn,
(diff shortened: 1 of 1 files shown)


src/composables/useFormChanged.mdx
```diff
@@ -0,0 +1,25 @@
...
+    additionalFields,
+    onCloseFn,
+    {mountBeforeUnload: true} // default: false, set to true if it should mount a beforeunload listener
+);
+
+// Can be used with modals or close handlers
+registerCloseCallback(() => {
+    confirmClose();
+
+    return !hasStateChanged();
```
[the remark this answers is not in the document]
blesildaramirez: Hi Jarda, yes the `onCloseFn` is needed in this example, as it called the `confirmClose` function. If the user confirms the dialog, then it will call the `onCloseFn`.

I will add another example for when it's not required to show the warning dialog, then the `onCloseFn` will not be required.

src/composables/useFormChanged.mdx
```diff
@@ -0,0 +1,40 @@
...
+const { confirmClose } = useFormChanged(
+  form,
+  additionalFields,
+  onCloseFn,
+  { warnOnClose: true }
+);
+
+```
+
+You can then pass `confirmClose` as the @cancel handler in your Form component.
```
jardakotesovec, line 40: I see - so its not possible to rely on the cancel just to trigger close and than useFormChanged would detect it via that registerCloseCallback?

blesildaramirez: Hi Jarda, it's possible. I forgot to update the documentation for this, it should now reflect the change for the `@cancel` event. https://github.com/pkp/ui-library/pull/691/commits/a045ef457a1c77b5b70a14ef962aee1bc6bc230c

PR #702, opened 2025-09-11
pkp/pkp-lib#11327 Public comment headless version

jardakotesovec:


taslangraham (approved):
This all looks good now @jardakotesovec.

I like that these headless components makes it easy to rearrange content, and style things without touching the components themself. My only gripe is the sheer number of components 😅, but that's a price worth paying for the flexibility.

For the outstanding things to consider, namely:

> order of the versions is incorrect right now, but I think you already have some solution for that

☝️  I can get a fix in today 

> we will need to figure out some mechanism to indicate that something is being processed

☝️  I might be able to work on this today.


> On Accessibility, there is probably room for improvement on use of semantic elements to express the comment section.

☝️  Willing to help in exploring this.

PR #704, opened 2025-09-16
structured citations

GaziYucel:


jardakotesovec:
For consistency I think this button styling will be better on dialog (secondary and warnable).. checked with couple of dialogs that Devika created when there is something to confirm:

Screenshot 2025-09-17 at 11 26 48
Screenshot 2025-09-17 at 11 25 58

src/managers/CitationManager/CitationManagerCellCitation.vue
```diff
@@ -0,0 +1,125 @@
...
+			</div>
+			<!-- is structured -->
+			<div v-if="citation.isStructured">
+				<div>{{ citation.title }}</div>
+				<div v-if="toggleStatusRow">
+					<div class="text-lg-normal leading-[2rem]">
+						<span v-for="(author, authorIndex) in citation.authors" :key="authorIndex">
+							<span>{{ author.familyName }} {{ author.givenName }}</span>
+							<a v-if="author.orcid" :href="author.orcid" target="_blank">
+								<Icon
```
jardakotesovec, line 33: It works.. but offsetting things by px to align items is not ideal..  I think inline-flex is to rescue here:

```
						<span
							v-for="(author, authorIndex) in citation.authors"
							:key="authorIndex"
							class="inline-flex items-center"
						>
							<span>{{ author.familyName }} {{ author.givenName }}</span>
							<a
								v-if="author.orcid"
								class="inline-flex items-center"
								:href="author.orcid"
								target="_blank"
							>
								<Icon icon="Orcid" class="mx-1 w-4" />
							</a>
						</span>

```

src/managers/CitationManager/CitationManagerCellCitation.vue
```diff
@@ -0,0 +1,125 @@
...
+				<span v-if="citation.urn">urn: {{ citation.urn }}</span>
+			</div>
+			<!-- is structured -->
+			<div v-if="citation.isStructured">
+				<div>{{ citation.title }}</div>
+				<div v-if="toggleStatusRow">
+					<div class="text-lg-normal leading-[2rem]">
+						<span v-for="(author, authorIndex) in citation.authors" :key="authorIndex">
+							<span>{{ author.familyName }} {{ author.givenName }}</span>
+							<a v-if="author.orcid" :href="author.orcid" target="_blank">
```
jardakotesovec, line 32: Having link just around icon is not great for accessibility. Icon is not useful for screen reader. 

So I would suggest something like:

```
<!-- Icon component, marked as decorative -->
    <Icon icon="Orcid" class="mx-1 w-4" aria-hidden="true" />
    <!-- Screen reader-only text with author context -->
    <span class="sr-only">ORCID profile for {{ author.givenName }} {{ author.familyName }}</span>
```

src/managers/CitationManager/CitationManager.vue
```diff
@@ -0,0 +1,105 @@
+<template>
+  <CitationManagerMetadataLookup/>
+  <CitationManagerAddRawCitations/>
+  <CitationManagerStatusProcessed
+      v-if="citationStore.citationsMetadataLookup"
+  />
+  <div>
+    <a
```
jardakotesovec, line 8: Lets use the button.. we do have link like buttons:

https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/story/components-button--link-like-button

src/managers/CitationManager/CitationManagerMetadataLookup.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+	<div>
+		<div class="pkpFormField__heading pkpFormFieldLabel">
+			{{ t('submission.citations.structured') }}
+		</div>
+		<div class="pkpFormField__description py-4">
+			{{ t('submission.citations.structured.citationsMetadataLookup.description') }}
+		</div>
+		<div>
+			<input
```
jardakotesovec, line 10: Can you use the Checkbox component. https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/story/components-checkbox--with-label

src/managers/CitationManager/CitationManagerMetadataLookup.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+	<div>
+		<div class="pkpFormField__heading pkpFormFieldLabel">
```
jardakotesovec, line 3: lets not use form field classes outside of the form, just use tailwind to make it consistent with rest of the page

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+	<div>
+		<div class="pkpFormField__heading pkpFormFieldLabel">
```
jardakotesovec, line 3: Same here, lets just use tailwind and semantic elements..   like h3

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
...
+		<div>
+			<textarea
+				class="pkpFormField__input pkpFormField--textarea__input !h-[9em]"
+				v-model="citationStore.citationsRawToBeAdded"
+			></textarea>
+		</div>
+		<div>
+			<PkpButton
+				class="my-2 mr-2"
+				:is-required="true"
```
jardakotesovec, line 18: I don't think we have this props on PkpButton

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
...
+		</div>
+		<div>
+			<textarea
+				class="pkpFormField__input pkpFormField--textarea__input !h-[9em]"
+				v-model="citationStore.citationsRawToBeAdded"
+			></textarea>
+		</div>
+		<div>
+			<PkpButton
+				class="my-2 mr-2"
```
jardakotesovec, line 17: Be careful when using left/right margin/padding, as that does not work well for RTL languages.

You can use ms-2 ms-3 ps-2 and pe-2  as these work with "start/end" instead of "left/right".

But I would suggest to simplify this..  and just make the button disabled when there is nothing in the textarea.. That should make obvious that something needs to be entered to add it..   

Otherwise we would have to announce correctly the feedback, so its also working for screen readers. So easier I think is just to simplify it.

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
...
+	<div>
+		<div class="pkpFormField__heading pkpFormFieldLabel">
+			{{ t('submission.citations') }}
+		</div>
+		<div class="pkpFormField__description py-4">
+			{{ t('submission.citations.structured.description') }}
+		</div>
+		<div>
+			<textarea
+				class="pkpFormField__input pkpFormField--textarea__input !h-[9em]"
```
jardakotesovec, line 11: Ideally we would have textarea component to apply some default styling, like we do have for checkbox component. We don't have that yet.. 

But in this particular case I am actually thinking that implementing this as super small form, using the useForm would be good fit. As you could indicate that the textfield is required for example.. and you would get correct semantic html for the description and textarea.

On main branch we got lot better in creating forms client side with `useForm`.

We currently support subset of fields there. I asked Blessie to add textfield (should be ready by tomorrow).

With that you can create form quickly client side..  you can see good examples in useDiscussionManagerForm.js from Blessie

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
...
+			<span
+				v-if="citationStore.citationsRawShowMessage === 'isEmpty'"
+				class="align-middle font-normal"
+			>
+				<Icon
+					:icon="'Declined'"
+					:class="'inline-block h-auto w-6 align-middle'"
+					:inline="true"
+				/>
+				<span class="align-middle font-normal">
```
jardakotesovec, line 32: Not sure what the font-normal would mean here?

You can pretty much control typography with these classes - https://main--6555d3db80418bb1681b8b17.chromatic.com/?path=/docs/guide-style-fonts--docs

src/managers/CitationManager/CitationManagerAddRawCitations.vue
```diff
@@ -0,0 +1,81 @@
...
+						t('submission.citations.structured.addRaw.empty')
+					}}
+				</span>
+			</span>
+			<span v-if="citationStore.citationsRawShowMessage === 'isLoading'">
+				<Spinner />
+			</span>
+			<span
+				v-if="citationStore.citationsRawShowMessage === 'isSuccess'"
+				class="items-center py-[0.5rem] text-success"
```
jardakotesovec, line 43: py with 0.5 rem is py-2, no need to use custom value

jardakotesovec (changes requested):
Its quite massive PR. I also did bunch of testing - and its much more stable than last time. So really happy about that.

I will take it for one more round of review after the feedback is incorporated. 

Thanks!

src/managers/CitationManager/citationManagerStore.js
```diff
@@ -0,0 +1,339 @@
...
+		 * add new raw citations
+		 */
+		const citationsRawToBeAdded = ref('');
+		const citationsRawShowMessage = ref('');
+
+		async function handleAddCitationsRawToList() {
+			if (!citationsRawToBeAdded.value) {
+				citationsRawShowMessage.value = 'isEmpty';
+				setTimeout(() => {
+					citationsRawShowMessage.value = 'isEmpty';
```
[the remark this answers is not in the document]
GaziYucel: This line `setTimeout(() => { citationsRawShowMessage.value = 'isEmpty'; ...` should actually be `setTimeout(() => {citationsRawShowMessage.value = ''; ...`

Is corrected.

src/managers/CitationManager/CitationManagerCellCitation.vue
```diff
@@ -0,0 +1,125 @@
...
+									:inline="true"
+								/>
+							</a>
+						</span>
+					</div>
+					<div v-if="citation.sourceName">
+						{{ citation.sourceName }}
+					</div>
+					<div class="leading-[2rem]">
+						<span v-if="citation.date">Date: {{ citation.date }}</span>
```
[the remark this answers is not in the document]
GaziYucel: Do you have an example of how this would look like?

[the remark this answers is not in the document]
GaziYucel: I have found an example and implemented this at the PHP side for the date. The date is only formatted in the list, in the edit form modal it's shown as Y-m-d.

I couldn't find an example for Volume.

PR #707, opened 2025-09-20
pkp/pkp-lib#11788 [stable-3_5_0] [TinyMCE] | Source Code text area - Journal Settings > Setup > Masthead > About the Journal, Selecting the source code option for tinyMCE displays shrunken text area

blesildaramirez:
- Since upgrading TinyMCE from [v5 to v7](https://github.com/pkp/pkp-lib/issues/9366), the CSS files were not correctly configured to use the proper version for TinyMCE v7. This was caused by the self-hosted TinyMCE CSS files in the lib/ui-library/public/styles/tinymce/ folder. In this PR ([#5101](https://github.com/pkp/ojs/pull/5101)), the app’s Vue config has been updated to load the correct styles from `node_modules`.
- The deleted files are not used anymore since we're directly using the correct source.


taslangraham:
Latest changes looks good to me @blesildaramirez

PR #709, opened 2025-09-25
pkp/pkp-lib#11825 Implement API integration for headnote and adding messages

blesildaramirez:


src/managers/DiscussionManager/DiscussionManagerTaskInfo.vue
```diff
@@ -113,6 +113,15 @@ watch(statusUpdateValue, (val) => {
 	emit('updateStatusCheckbox', val);
 });
 
+// update the checkbox state if statusValue from props changes
+watch(
+	() => props.statusValue,
+	(val) => {
```
jardakotesovec, line 119: So maybe props.statusValue could be used directly to avoid synchronisation here?

jardakotesovec (approved):
Just minor suggestion. Assuming that getting participant is to be addressed in next PR? As the participantManagerStore won't be available on author workflow page neither in reviewer wizard

src/managers/DiscussionManager/DiscussionManagerTaskInfo.vue
blesildaramirez: I didn't use the `props.statusValue` directly to avoid mutation error when the checkbox changes from the child component. I can explore using `defineModel` instead, I think it simplifies the implementation and works the same.
But for now I will merge this PR as you have approved it already. 😄

PR #794, opened 2026-02-06
pkp/pkp-lib#12262 [Frontend] Publication – Media Files

blesildaramirez:


src/components/Form/fields/FieldBase.vue
```diff
@@ -42,6 +42,8 @@ export default {
 		/** The `name` of another field which should have a truthy value before this field is shown.
 		 * You can also pass an array to require a specific value or multiple values. For example, `['primaryLocale', 'en_US']` would hide this field unless the `primaryLocale` field had a value of `en_US`. Or `['primaryLocale', ['en_US', 'fr_CA']]` to check for multiple values. */
 		showWhen: [String, Array],
+		/** Whether to hide this field when the form is rendered in a read-only/display-only (non-editable) mode, while keeping it visible in normal editable mode. */
+		hideOnDisplay: Boolean,
```
jardakotesovec, line 46: Is this addition specific for media section?

src/managers/MediaFileManager/mediaFileManagerStore.js
```diff
@@ -0,0 +1,228 @@
...
+			`submissions/${submission.value.id}/publications/${publication.value.id}/mediaFiles`,
+		);
+
+		const {
+			items: mediaFilesList,
+			fetch: fetchMediaFiles,
+			isLoading: isLoadingMediaFiles,
+		} = useFetchPaginated(mediaFileApiUrl, {
+			page: 1,
+			pageSize: 999,
```
jardakotesovec: We don't handle pagination right? In that case I would be tempted to bump it at least to 9999 to be safe :-).

src/components/FileUploader/FileUploader.vue
```diff
@@ -21,6 +21,8 @@
 
 <script>
 import VueDropzone from 'dropzone-vue3';
+import {useDropzoneDragDrop} from '@/composables/useDropzoneDragDrop';
+import {parseDropzoneError} from '@/utils/fileUtils';
```
jardakotesovec: If these are both handling the dropzone - would not be good home for parseDropzoneError within useDropzoneDragDrop? :-)

src/utils/fileUtils.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Format a file size in bytes to a human-readable string
+ *
+ * @param {number} bytes - The file size in bytes
+ * @returns {string} Formatted size string (e.g., "1.5 MB")
+ */
+export function formatFileSize(bytes) {
```
jardakotesovec, line 7: Just for consistency thinking whether this would deserve its own  useFile, similarly like we have useDate? In general putting these things to composables for better discoverability. utils are usually just internal supporting files

src/components/InlineSelect/InlineSelect.vue
```diff
@@ -0,0 +1,96 @@
+<template>
```
jardakotesovec, line 1: This might be best to discuss on call, but I would be basically asking if there is we need to introduce borderless select.. or whether it could be just variant of the our usual FieldSelect.
I am sure you considered that option - so would like to learn more about the decision.

jardakotesovec (changes requested):
All very minor suggestions. Looks very well organised. Overall outstanding work on this! 

One thing I would like to learn more about is the choice of creating FieldSelectBorderLess.

src/components/Table/TableBodyGroup.vue
```diff
@@ -0,0 +1,42 @@
...
+
+const unregisterGroupRow = (id) => {
+	const idx = rowIds.value.indexOf(id);
+	if (idx !== -1) rowIds.value.splice(idx, 1);
+};
+
+// Provide group context to child TableRows
+provide('rowSpanGroupContext', {
+	groupId: props.groupId,
+	get groupSize() {
```
jardakotesovec: just that this pattern looks unusual compared to others? I think as long as you ensure that the variable you are passing is reactive it should work fine.

src/components/Table/TableRowSpan.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
jardakotesovec: I wonder how feasible would be just to add rowspan option on the tableCell? That would be probably bit more straightforward as API to understand..  as its closer to actual html

src/components/Form/fields/FieldBase.vue
blesildaramirez: No, this is for the task & discussions work. I just forgot to add the prop to the FieldBase component where a prop can be used in general for all the field components.

src/managers/MediaFileManager/mediaFileManagerStore.js
blesildaramirez: Actually not setting the pageSize would not make difference right now because the mediaFiles doesn't paginate and just returns all the files. But in any case there will be pagination implemented in the api, it would be good that frontend still passes this param so it won't break anything.
As for the 999 value, it was also set for t&d, task templates. I updated them too to use 9999.

src/components/InlineSelect/InlineSelect.vue
blesildaramirez: Yeah, I went back and forth on this. I first tried making it a variant of `FieldSelect`, but it ended up with two render paths and mixed structures (form-framework vs standalone v-model).

There is mismatch here, the `FieldSelect` is tied to the form framework, while this is just a simple standalone widget for media files, which is used inside table cell.

So I kept it separate and renamed it: `FieldSelectBorderless` -> `InlineSelect` (now in `components/InlineSelect/`). The `Field*` prefix felt misleading since this isn't part of the form framework.

src/components/Table/TableBodyGroup.vue
blesildaramirez: Yeah, used `toRef` instead - keeps the reactivity without the getter pattern.

src/components/Table/TableRowSpan.vue
blesildaramirez: Yes this is possible and makes more sense. Initially this was meant as a wrapper around `TableCell` when used with rowspan (to calculate the group context, etc.), but that was moved to `TableBodyGroup` - so no need to keep this component separately. 
I removed this component now and used the `TableCell` component instead.

PR #813, opened 2026-02-28
pkp/pkp-lib#1232 Support Funder Data

ajnyga:
https://github.com/pkp/pkp-lib/issues/12392


src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+		</div>
+		<div class="pkpFormField__control">
+			<div v-if="currentValue && currentValue.name">
+				<div v-if="currentValue.ror">
+					<div class="flex items-center text-lg-semibold">
+						{{ localize(currentValue.name) }}
+						<a
+							:href="currentValue.ror"
+							class="ms-2"
+							target="_blank"
```
blesildaramirez, line 23: Please add `rel="noopener noreferrer"` when using `target="_blank"` to avoid potential security issues.

src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+						>
+							<Icon icon="ROR" class="w-6" :inline="true" />
+						</a>
+					</div>
+
+					<a
+						:href="currentValue.ror"
+						class="block text-lg-semibold"
+						target="_blank"
```
blesildaramirez, line 33: Same comment above:
Please add rel="noopener noreferrer" when using target="_blank" to avoid potential security issues.

src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+	</div>
+</template>
+
+<script setup>
+import { ref, computed, watch, useId, onBeforeUnmount } from 'vue';
+import { t } from '@/utils/i18n';
+import {useUrl} from '@/composables/useUrl';
+import {useFetch} from '@/composables/useFetch';
+import { localize } from '@/utils/i18n';
+import Button from '@/components/Button/Button.vue';
```
blesildaramirez: Let’s name this `PkpButton` to keep it consistent with the rest of the codebase.

src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+							size="large"
+							:is-required="locale === primaryLocale"
+							@change="(_, __, value) => updateFunderName(locale, value)"
+						/>
+					</div>
+				</div>
+				<div class="mt-3">
+					<Button isWarnable @click="removeFunder">
+						{{ t('common.delete') }}
+					</Button>
```
blesildaramirez: Let’s name this PkpButton to keep it consistent with the rest of the codebase. Also pass the prop as `is-warnable` (hyphenated)

```suggestion
					<PkpButton is-warnable @click="removeFunder">
						{{ t('common.delete') }}
					</PkpButton>
```

src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+				/>
+
+			</div>
+		</div>
+	</div>
+</template>
+
+<script setup>
+import { ref, computed, watch, useId, onBeforeUnmount } from 'vue';
+import { t } from '@/utils/i18n';
```
blesildaramirez: We can include the import of `localize` here.

```suggestion
import { t, localize } from '@/utils/i18n';
```

src/components/Form/fields/FieldFunder.vue
```diff
@@ -0,0 +1,210 @@
...
+import {useUrl} from '@/composables/useUrl';
+import {useFetch} from '@/composables/useFetch';
+import { localize } from '@/utils/i18n';
```
blesildaramirez: We can remove when https://github.com/pkp/ui-library/pull/813/changes#r3083932698 is applied

```suggestion
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+			</label>
+		</div>
+		<div :id="descriptionId" class="pkpFormField__description">
+			{{ props.description }}
+		</div>
+		<PkpTable :labelled-by="labelId" :described-by="descriptionId">
+			<TableHeader>
+				<TableColumn id="">{{ t('submission.funders.funder.grant.doi', []) }}</TableColumn>
+				<TableColumn id="">{{ t('submission.funders.funder.grant.number', []) }}</TableColumn>
+				<TableColumn id="">{{ t('submission.funders.funder.grant.name', []) }}</TableColumn>
```
blesildaramirez: Since this translation doesn’t use any variables, we can remove the second parameter for clarity. We can also remove the `id=""` since it's empty.

```suggestion
				<TableColumn>{{ t('submission.funders.funder.grant.doi') }}</TableColumn>
				<TableColumn>{{ t('submission.funders.funder.grant.number') }}</TableColumn>
				<TableColumn>{{ t('submission.funders.funder.grant.name') }}</TableColumn>
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+				<TableColumn id="" class="w-[100px]">&nbsp;</TableColumn>
+			</TableHeader>
+			<TableBody>
+				<TableRow v-for="(row, index) in currentValue" :key="index">
+					<TableCell>
+						<FieldText
+							:name="'grantDoi'"
```
blesildaramirez: Since this is a static string, it can be set directly with:

```suggestion
							name="grantDoi"
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+							@change="
+								(fieldName, _, fieldValue) => {
+									updateRow(index, fieldName, fieldValue);
+								}
+							"
+						/>
+					</TableCell>					
+					<TableCell>
+						<FieldText
+							:name="'grantNumber'"
```
blesildaramirez: Since this is a static string, it can be set directly with:

```suggestion
							name="grantNumber"
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+							@change="
+								(fieldName, _, fieldValue) => {
+									updateRow(index, fieldName, fieldValue);
+								}
+							"
+						/>
+					</TableCell>
+					<TableCell>
+						<FieldText
+							:name="'grantName'"
```
blesildaramirez: Same comment as https://github.com/pkp/ui-library/pull/813/changes#r3083994536:
Since this is a static string, it can be set directly with:

```suggestion
							name="grantName"
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+							"
+						/>
+					</TableCell>
+					<TableCell>
+						<PkpButton
+							:is-link="true"
+							:is-warnable="true"
+							@click="deleteRow(index)"
+						>
+							{{ t('common.delete', []) }}
```
blesildaramirez: We can remove the 2nd param as this translation doesn't expect a variable for clarity.

```suggestion
							{{ t('common.delete') }}
```

src/components/Form/fields/FieldFunderGrants.vue
```diff
@@ -0,0 +1,171 @@
...
+	formId: {
+		type: String,
+		default: null,
+	},
+	/** The `<label>` for this field. May be used in a `<fieldset>` when appropriate. All form fields should have an accessible label. */
+	label: {
+		type: String,
+		default: null,
+	},
+	/** Adds a description to the field. Can include HTML code. */
```
blesildaramirez: I think if `description` contains HTML, it won't be rendered correctly with `{{ props.description }}` and may require `v-strip-unsafe-html` instead (https://github.com/pkp/ui-library/pull/813/changes#diff-2ae5c97c701828a6941e61908c52df48e869cda6b9c05ae90406d92160b714ddR12)

We can use `v-strip-unsafe-html` and keep this comment, but I don't see a use case for it so I suggest removing the comment that says it can include HTML code:
```suggestion
	/** Adds a description to the field. */
```

src/managers/FunderManager/modals/FunderEditModal.vue
```diff
@@ -0,0 +1,35 @@
...
+					@set="(...args) => emit('set', ...args)"
+					@success="
+						(...args) => {
+							closeModal();
+							emit('success', ...args);
+						}
+					"
+				></PkpForm>
+			</div>
+		</div>
```
blesildaramirez: I don't see any styling for this class, we can remove this. Alternatively, we could use the `SideModalLayoutBasic` to standardize the form layout inside the modal.

```suggestion
		<SideModalLayoutBasic>
			<PkpForm
				v-bind="form"
				@set="(...args) => emit('set', ...args)"
				@success="
					(...args) => {
						closeModal();
						emit('success', ...args);
					}
				"
			></PkpForm>
		</SideModalLayoutBasic>
```

src/managers/FunderManager/modals/FunderEditModal.vue
```diff
@@ -0,0 +1,35 @@
...
+	</SideModalBody>
+</template>
+
+<script setup>
+import {inject, defineEmits} from 'vue';
+import SideModalBody from '@/components/Modal/SideModalBody.vue';
```
blesildaramirez, line 23: If using `SideModalLayoutBasic `, don't forget to import:

```suggestion
import SideModalBody from '@/components/Modal/SideModalBody.vue';
import SideModalLayoutBasic from '@/components/Modal/SideModalLayoutBasic.vue';
```

src/managers/FunderManager/FunderManager.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+	<PkpTable :aria-label="t('submission.funders')">
+		<template #label>
+			<h3 class="text-4 font-semibold">
```
blesildaramirez: I see that different styles are used for table labels across the codebase. We don't seem to have standard font class to use, so it's either based from a Figma design or maybe just what looks appropriate when testing. Sometimes it's `text-xl-bold` or `text-3xl-bold`, sometimes it's not set for other manager components.

But in this case, the `text-4` is not a valid class from our Tailwind config - we should remove it. As for the font class to use, I think it's better to use `text-xl-bold`, this is the comparison from both classes:

Image

src/managers/FunderManager/FunderManager.vue
```diff
@@ -0,0 +1,79 @@
...
+				{{ t('submission.funders') }}
+			</h3>
+			<span class="text-lg-normal">
+				{{ t('submission.funders.description') }}
+			</span>
+		</template>
```
blesildaramirez: On PkpTable component, there is a dedicated slot for description, I think we should use it here:
(used `<p>` here to align use with other manager components)

```suggestion
		</template>
		<template #description>
			<p class="text-lg-normal">
				{{ t('submission.funders.description') }}
			</p>
		</template>
```

src/managers/FunderManager/FunderManagerCellFunder.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+	<TableCell :is-row-header="false">
+		<div>
+			<div>
+				<div class="text-lg-normal">
+					{{ localize(funder.name) }}
+					<Icon v-if="funder.ror" icon="ROR" class="w-6" :inline="true" />
+				</div>
+			</div>
+		</div>
```
blesildaramirez: We can remove the redundant nested `<div>` elements here to simplify the structure.

```suggestion
		<div class="text-lg-normal">
			{{ localize(funder.name) }}
			<Icon v-if="funder.ror" icon="ROR" class="w-6" :inline="true" />
		</div>
```

blesildaramirez (changes requested):
Hi @ajnyga , I've reviewed the frontend changes and left a few suggestions, mostly around standardizing the implementation. Please let me know if you have any questions or need any clarification. Thank you!

src/managers/FunderManager/FunderManager.vue
ajnyga: I think this came from the Data Citations implementation, very few examples existed when I made that and do not recall what example I ended up using.

The font size you chose looks better to my eye as well.

ajnyga:
> Hi @ajnyga , I've reviewed the frontend changes and left a few suggestions, mostly around standardizing the implementation. Please let me know if you have any questions or need any clarification. Thank you!

Thanks for the review @blesildaramirez, I've taken all your comments into account.

blesildaramirez:
Hi @ajnyga , just checking if you were able to push the changes. I'm not seeing the new commits or updates on my end yet. Thanks!

ajnyga:
Hi, @blesildaramirez was just waiting for an unresolved question in pkp-lib before pushing and starting the tests

kaitlinnewson:
I'm not sure if it was discussed yet, but should the button ordering be changed here, with "Order" on the far left, to be similar to the contributors panel, and also have the arrows on the button? I think this also applies to Data Citations. I've also suggested a locale key change in `pkp-lib` to align the UI text.

**Contributors panel:**

Screenshot 2026-05-05 at 10 44 10 AM

**Current funding panel:**

Screenshot 2026-05-05 at 10 43 36 AM

ajnyga:
The component used in Funders is a Manager, so not entirely the same as the panel component used with contributors.

But let me know if we want to align these like @kaitlinnewson  suggests.

blesildaramirez (approved):
Hi @ajnyga , thanks for addressing all my feedback - approved! However, please switch the ordering of the buttons as suggested by Kaitlin, so it'd be "Order" -> "Add a new Funder". The reasoning is that primary actions usually sit on the far right end, so the eye lands there last.
Also, adding the "Sort" icon would help - makes it clearer the button is for sorting. 

Please feel free to merge once you've addressed the couple of items above. Thanks!

Also, thanks @kaitlinnewson - good catch on these!

ajnyga:
Thanks, I will change the order and add the icon. 

@blesildaramirez also @kaitlinnewson raised the suggestion to shorten the label to "Add Funder".

The longer label comes from UI mockups that I received, but Kaitlin had a good point that in many other places labels like these are shorter ("Add Galley", "Add Contributor")

So I am adding that change as well. I can revert later if you think it should be the longer label.

blesildaramirez:
Hi @ajnyga , that makes perfect sense - please proceed with the updated label for the "Add Funder". Thank you!

PR #853, opened 2026-03-25
Optimize table section reloads after closing an unsaved modal

blesildaramirez:
**Details of the changes:**
* Side modals now track their own `dataChanged` state and only reload the table when something changed. This applies regardless of how the modal is closed. Changes from nested modals are also propagated to the parent modal.
* * File Uploads and Add Galley still always reload, since the legacy upload flow has multiple steps and a file can be uploaded before `dataChanged` is passed at the final step. The submissions list also always reloads when the workflow modal closes, to ensure up-to-date submission list in the dashboard.
* Some sections don't use `triggerDataChange()` and refetch their own lists (e.g `src/pages/userComments/userCommentStore.js`), so they check the close data with `shouldTriggerDataChange(closeData)`.
* Removed wrapper functions around `triggerDataChange()` so the close data passes through directly.
* Closing a side modal now closes the modal first, then shows a full-screen freeze while the table reloads. `onClose` is awaitable so the freeze stays up until the reload is complete.
* Fixed a few double reloads found while testing.
* Added documentation for the new behaviour and a unit test for the modal store.

Local testing:
I went through every area the PR touches in the browser and the tests went fine: galleys, media, files, participants, references, reviewers, users, task templates, comments, discussions, and author response.

### DIFF
--- a/src/composables/useFetch.js
+++ b/src/composables/useFetch.js
@@ -150,7 +150,7 @@ export function useFetch(url, options = {}) {
 
 			// Auto-track data changes for mutations inside modals
 			if (
-				['POST', 'PUT'].includes(options.method) &&
+				['POST', 'PUT', 'DELETE'].includes(options.method) &&
 				modalStore.dialogLevel > 0
 			) {
 				modalStore.markModalDataChanged(modalStore.dialogLevel);
--- a/src/composables/useFetch.js
+++ b/src/composables/useFetch.js
@@ -151,9 +151,9 @@ export function useFetch(url, options = {}) {
 			// Auto-track data changes for mutations inside modals
 			if (
 				['POST', 'PUT', 'DELETE'].includes(options.method) &&
-				modalStore.dialogLevel > 0
+				modalStore.modalLevel > 0
 			) {
-				modalStore.markModalDataChanged(modalStore.dialogLevel);
+				modalStore.markModalDataChanged(modalStore.modalLevel);
 			}
 		} catch (e) {
 			if (signal) {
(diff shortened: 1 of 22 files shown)


src/composables/useFetch.js
```diff
@@ -147,6 +147,14 @@ export function useFetch(url, options = {}) {
 			data.value = result;
 			validationError.value = null;
 			isSuccess.value = true;
+
+			// Auto-track data changes for mutations inside modals
+			if (
+				['POST', 'PUT', 'DELETE'].includes(options.method) &&
+				modalStore.dialogLevel > 0
```
jardakotesovec: should be used the modalLevel which is available above?

Might not make difference but dialogLevel was made intentionally to guide dialogs on which levels to show up.

jardakotesovec (changes requested):
Lets discuss this further on catch up. One of the goals I think was to reload the data before closing the modal, which per my testing is not happening (tried it on the task template manager).
Otherwise conceptually I think its good idea.

jardakotesovec (approved):
@blesildaramirez I think its ready to be shipped! I tried to also improve the animation - so the loader kicks in after the side modal finishes - which in many cases is done before side modal animation finishes so it does not show up at all. So if you agree, add these changes on top of yours and we can spin up e2e-bot to get ready for this - https://github.com/jardakotesovec/ui-library/commit/bc43bd4a5d571dc3cb6872c394358950208d29f2

blesildaramirez:
Hi @jardakotesovec , thanks for the review! I cherry-picked your commit to include it to this PR's changes.
Tests are currently running - please feel free to merge after that or when e2e is ready to accommodate the changes. Thanks!

PR #864, opened 2026-04-07
pkp/pkp-lib#12531 Show tasks and discussion activities by categories

blesildaramirez:


src/composables/useDate.js
```diff
@@ -88,5 +89,13 @@ export function useDate() {
...
 		 */
 		getRelativeTargetDate,
+
+		/**
+		 * Check if a date is within the next 7 days
+		 * @type {Function}
+		 * @param {Date|string} date - The date to check
+		 * @returns {boolean} True if the date is within the next 7 days, false otherwise
+		 */
+		isWithin7Days,
```
jardakotesovec: Good idea to add util for this, maybe just to make it bit more universal - what about `isWithinDays` and the number would be argument? So its `isWithinDays(7)`

src/managers/DiscussionManager/DiscussionManagerCellActivity.vue
```diff
@@ -1,15 +1,47 @@
...
 <script setup>
+import {computed} from 'vue';
+import {useDate} from '@/composables/useDate';
 import TableCell from '@/components/Table/TableCell.vue';
 
-defineProps({workItem: {type: Object, required: true}});
+const props = defineProps({workItem: {type: Object, required: true}});
+const {isWithin7Days} = useDate();
+
+const recentNote = computed(() => {
```
jardakotesovec: Looks great, maybe just to make it even more readable, could be named something like recentNotePostedActivity

jardakotesovec (approved):
Looks good, small suggestions, after that can be merged.

PR #868, opened 2026-04-09
pkp/pkp-lib#12497 Restrict access to edit participant button

Vitaliy-1:


src/managers/ParticipantManager/participantManagerStore.js
```diff
@@ -56,6 +57,12 @@ export const useParticipantManagerStore = defineComponentStore(
 						recommendOnly: stageAssignment.recommendOnly,
 						displayInitials: participant.displayInitials,
 						canLoginAs: participant.canLoginAs,
+						isEditable: canCurrentUserEditParticipant(
+							participant,
+							stageAssignment,
+							props.submission,
+							participants
+						),
```
jardakotesovec: This overall looks great. My only preference would be to have this calculation in getItemActions. 

Can you try - I think it should be quite easy change - if you need more data available in the getItemActions. You can add them

```
		function getItemActions({participant}) {
			return participantManagerConfig.getItemActions({
				submission: props.submission,
				submissionStageId: props.submissionStageId,
				participant,
			});
		}

```

Reason for it that this place is just for mapping to make the participant shape easier to work with. But the logic what to display is happening in the config functions.

jardakotesovec (commented):
Provided one suggestion on organisation if you can try. But its not deal breaker :-).

PR #876, opened 2026-04-14
pkp/pkp-lib#12155 only show processing status for structured citations

kaitlinnewson:


blesildaramirez (approved):
Looks good on my end!

PR #879, opened 2026-04-17
pkp/pkp-lib#12337 CSS masking to handle auto complete off for password type input

touhidurabir:
for pkp/pkp-lib#12337 .

### DIFF
--- a/src/components/Form/fields/FieldText.vue
+++ b/src/components/Form/fields/FieldText.vue
@@ -36,14 +36,16 @@
 					ref="input"
 					v-model="currentValue"
 					class="pkpFormField__input pkpFormField--text__input"
-					:type="inputType"
+					:class="{'pkpFormField--text__input--masked': isMaskedSecret}"
+					:type="effectiveInputType"
 					:name="localizedName"
 					:aria-describedby="describedByIds"
 					:aria-invalid="!!errors?.length"
 					:disabled="isDisabled"
 					:required="isRequired"
 					:style="inputStyles"
-					:autocomplete="autocomplete"
+					:autocomplete="effectiveAutocomplete"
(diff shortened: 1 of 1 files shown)


src/components/Form/fields/FieldText.vue
```diff
@@ -160,6 +177,85 @@ export default {
...
+		 * When true, the input renders as type="text" with CSS
+		 * `-webkit-text-security: disc` instead of type="password", so
+		 * browsers don't classify the form as a login form and don't
+		 * offer saved-login autofill.
+		 *
+		 * When the browser does NOT support CSS masking (Firefox < 114),
+		 * this returns false and the field falls back to type="password"
+		 * with autocomplete="new-password" + vendor data-* attrs.
+		 */
+		isMaskedSecret() {
```
jardakotesovec, line 200: Hi @touhidurabir , seems that the different browsers and extensions are being quite inconsistent on this. 

I will try to push back here (sorry), to see whether we could go with simpler version, as this is quite convoluted. 

Could not we for compliance purposes just go with autocomplete=off and if the extensions browsers fails to follow that - I don't think thats our problem, but they are not following attribute which is meant to achieve this behaviour. 

@asmecher Any thoughts on how far we need to take this?

touhidurabir: > this is quite convoluted.

I agree on this 100% . 

> Could not we for compliance purposes just go with autocomplete=off and if the extensions browsers fails to follow that - I don't think thats our problem, but they are not following attribute which is meant to achieve this behaviour.

If we can do that, that's best for us . However European Commission Web Application Security Standard set this as a requirement . And seems some browser and extension just simply does not respect the attr `autocomplete=off`, specially when considering FireFox only as browser . So I make an update PR on this to handle such cases .

PR #880, opened 2026-04-19
pkp/pkp-lib#11825 Fix form visible locales for task and discussion forms

blesildaramirez:


src/managers/DiscussionManager/useDiscussionManagerForm.js
```diff
@@ -590,6 +592,7 @@ export function useDiscussionManagerForm(
 	initEmptyForm(formId, {
 		showErrorFooter: true,
 		canSubmit: !inDisplayMode,
+		locales: [getCurrentLocale()],
```
jardakotesovec: Sorry, I did suggest solution to set the locales here. But when I had bit more time to think about. The form is not multilingual simply because it does not have any multilingual field. 

And it seems to be considered in Form.vue
```
		availableLocales() {
			if (!this.hasMultilingualFields) return [];
			return [...this.supportedFormLocales].sort((a, b) => {
				if (a.key === this.primaryLocale) return -1;
				if (b.key === this.primaryLocale) return 1;
				return 0;
			});
		},

```
So maybe this also could be addressed on displaying side of things - to basically make sure that if there is no multilingual field - it would not jump to partially multilingual view? As ideally for such form you would not need to think about locales at all.

blesildaramirez: Hi Jarda, you're right. The main issue here is that the class `pkpForm--hasManyVisibleLocales` is being applied if there are more than one `visibleLocales`. I think it needs to also check if the form `hasMultilingualFields`. Could you check the latest changes for this PR? Thanks!

jardakotesovec (approved):
Great, thanks!

PR #883, opened 2026-04-23
pkp/pkp-lib#12509 Add Crossref support for peer review DOIs

taslangraham:
For pkp/pkp-lib#12509


src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -100,6 +104,33 @@
...
+
+						<TableCell>
+							<!-- Status of individual DOI objects -->
+							<Badge
+								class="doiListItem__itemMetadata--badge"
+								:is-warnable="
+									(!isDeposited[row.id] && isRegistrationPluginConfigured) ||
+									(isDeposited[row.id] &&
+										isRegistrationPluginConfigured &&
+										isStale(row.itemDepositStatus)) ||
```
bozana: `row.depositStatus` instead of `row.itemDepositStatus` ? -- row is an iteration variable over item.doiObjects

bozana (commented):
Hi @taslangraham, just a comment/question. I am unfortunately not the right person to review UI 😅

src/components/ListPanel/doi/DoiListItem.vue
taslangraham: You're right. Thanks for catching this.

taslangraham:
Thanks for the review @bozana. @blesildaramirez Do you mind giving this a second review, please? Details of the UI changes are listed here: https://github.com/pkp/pkp-lib/issues/12509#issuecomment-4417000625. Thanks in advance.

src/components/ListPanel/doi/useDoiComposable.js
```diff
@@ -0,0 +1,130 @@
+// useDoiComposable
+
+import {computed, toRef} from 'vue';
+import {useLocalize} from '@/composables/useLocalize';
+
+export function useDoiComposable(props) {
```
blesildaramirez: Our composable files don't usually include the word Composable, could we rename this with just `useDoi`? Thanks!

src/components/ListPanel/doi/useDoiComposable.js
```diff
@@ -0,0 +1,130 @@
...
+			(doiObject) => doiObject.isCurrentVersion,
+		);
+	});
+
+	/**
+	 * Deposit status for top-level item
+	 */
+	const itemDepositStatus = computed(() => {
+		return currentVersionDoiObjects.value.length !== 0
+			? currentVersionDoiObjects.value[0].depositStatus
```
blesildaramirez: I think we should use optional chaining here to safely access the property:

```suggestion
			? currentVersionDoiObjects?.value?.[0]?.depositStatus
```

src/components/ListPanel/doi/useDoiComposable.js
```diff
@@ -0,0 +1,130 @@
...
+		const depositedStatuses = [
+			pkp.const.DOI_STATUS_SUBMITTED,
+			pkp.const.DOI_STATUS_REGISTERED,
+		];
+
+		const results = {
+			[item.value.id]: depositedStatuses.includes(itemDepositStatus.value),
+		};
+
+		item.value.doiObjects.forEach((doiObject) => {
```
blesildaramirez: I'm confused as to why we needed to build an `id-keyed` maps here, I think we can simply drop it and just make this a function if I'm understanding it right?

Each call site already has the status in hand (itemDepositStatus / row.depositStatus), so keying by id doesn't seem to buy us anything. Could we try if we can use the function instead (to be consistent with `isStale` and `getDepositStatusString`)

So this computed becomes:
```
function isDeposited(status) {
      return [
          pkp.const.DOI_STATUS_SUBMITTED,
          pkp.const.DOI_STATUS_REGISTERED,
      ].includes(status);
  }
```

src/components/ListPanel/doi/useDoiComposable.js
```diff
@@ -0,0 +1,130 @@
...
+
+	/**
+	 * Error states keyed by object id
+	 */
+	const hasErrors = computed(() => {
+		const results = {
+			[item.value.id]: itemDepositStatus.value === pkp.const.DOI_STATUS_ERROR,
+		};
+
+		item.value.doiObjects.forEach((doiObject) => {
```
blesildaramirez: Same here, I think we can drop the mapping and can just check if the status (when added as param) can just check if it matches DOI_STATUS_ERROR.

```
function hasError(status) {                                                                                                                                                                                         
        return status === pkp.const.DOI_STATUS_ERROR;                                                                                                                                                                   
    } 
```

From what I can see in the current usages of `hasErrors`, the mapping doesn't appear to be needed. Could you help me understand the use case for it if it is required?

blesildaramirez (changes requested):
Hi @taslangraham , done reviewing. My suggestions are mostly for clarification - whether the `id-keyed` objects are necessary for the 2 computed properties - plus a small naming nit for the `useDoi` composable and a note on using optional chaining for the nullable fields. Thanks!

src/components/ListPanel/doi/useDoiComposable.js
taslangraham: Having reviewed this, you're right, keying by ID is not necessary here. This and the`hasErrorrs` computed property can be converted to a function. Thanks.

taslangraham:
@blesildaramirez, thanks for the review. I've made updates based on your comments, and this is now ready for a second round of review. The commit with changes based on your review can be found here: https://github.com/pkp/ui-library/pull/883/changes/9a71271fe7bfb57c2a5258de00c1c01a272fc5f2.

blesildaramirez (approved):
Thanks @taslangraham , this is all good now.

src/components/ListPanel/doi/DoiListItem.vue
```diff
@@ -39,17 +39,22 @@
 				{{ item.id }}
 				<div class="doiListItem__itemMetadata">
 					<Badge
-						class="doiListItem__itemMetadata--badge"
+						:is-primary="item.isPublished && isDeposited(item.depositStatus)"
```
bozana: Item is the mapped item from DoiListPanel.mappedItems, which has no depositStatus property (only entries inside item.doiObjects[] have it).
So item.depositStatus is always undefined here and everywhere below it (lines 45, 47, 50, 169, 187, 203, 211, 218, 781). All those calls should use itemDepositStatus instead — the computed from useDoi which correctly derives it from currentVersionDoiObjects[0].depositStatus. 
isStale(itemDepositStatus) at line 49 is correct.

src/components/ListPanel/doi/DoiListPanel.vue
```diff
@@ -573,15 +573,19 @@ export default {
 					},
 					/**
 					 *
-					 * @param {{temporaryFileId: int}} data
+					 * @param {{temporaryFileIds: int}} data
```
bozana: `@param {{temporaryFileIds: int[]}} data` -- it is an array now

src/components/ListPanel/doi/useDoi.js
```diff
@@ -0,0 +1,111 @@
+// useDoi
```
bozana: no need for this?

bozana (commented):
Hi @taslangraham, this is what Claude find out :-)

PR #885, opened 2026-04-29
pkp/pkp-lib#12584 Add "Summary of Changes/Amendment" field on file revision upload and update publication issues page

blesildaramirez:


src/managers/FileManager/modals/FileMetadataForm.vue
```diff
@@ -0,0 +1,76 @@
+<template>
```
jardakotesovec, line 1: Suggesting to move this somewhere under the FileManager.. as its eventually all the modals being opened from FileManager will be there..  this is some interment building block..  but I think its good to see there that this is one of the modern pieces of FileManager

src/managers/FileManager/modals/useFileMetadataForm.js
```diff
@@ -0,0 +1,179 @@
...
+		groupId: 'default',
+		label: t('submission.form.name'),
+		isMultilingual: true,
+		isRequired: true,
+		size: 'large',
+		value: submissionFile.name || {},
+	});
+
+	if (submissionFile.fileStage === pkp.const.SUBMISSION_FILE_REVIEW_REVISION) {
+		addFieldRichTextArea('summaryOfChanges', {
```
jardakotesovec, line 107: Don't we have same logic for Media FIles? Just wondering whether that will eventually be using the same 'form'?

jardakotesovec (commented):
Looks great, lets discuss whether we could incorporate the Update Type within this work.

src/managers/FileManager/modals/useFileMetadataForm.js
blesildaramirez: Hi @jardakotesovec , yes this could be used for Media Files as well. Although might needs adjustment with the api communication when saving, as the current implementation for this PR is it uses the legacy save form. For Media Files, there is work to add save metadata api but that's not yet merged to main. This will need to be updated again when that api is available.

blesildaramirez: With that, I will rename the `SubmissionFileMetadataForm.vue` with just `FileMetadataForm.vue`

src/composables/useInsertSummaryOfChangesContent.js
```diff
@@ -0,0 +1,73 @@
...
+						if (!editor) return;
+						// Collapse selection to the end so insertContent appends.
+						const range = editor.dom.createRng();
+						range.selectNodeContents(editor.getBody());
+						range.collapse(false);
+						editor.selection.setRng(range);
+						editor.insertContent(html);
+					},
+				);
+				window.scrollTo({top: pageScroll, behavior: 'instant'});
```
jardakotesovec: Why the scroll here? While testing it jumps to the top after inserting - away from the Summary of Changes field, which I don't think its helpful.

src/components/InsertSummaryOfChanges/InsertSummaryOfChangesModal.vue
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const fileStages = [pkp.const.SUBMISSION_FILE_REVIEW_REVISION];
+if (isOMP()) {
+	fileStages.push(pkp.const.SUBMISSION_FILE_INTERNAL_REVIEW_REVISION);
+}
+const {apiUrl} = useUrl(`submissions/${props.submissionId}/files`, {
+	fileStages: fileStages.join(','),
+});
+const {data, fetch, isLoading} = useFetch(apiUrl);
```
jardakotesovec: Always helpful to use some specific names with the syntax of {data: files, fetch: fetchFiles, ...}, it could be `files` or `filesData`

src/composables/useInsertSummaryOfChangesContent.js
```diff
@@ -0,0 +1,73 @@
...
+				});
+			}
+		},
+	};
+
+	function openInsertModal() {
+		openSideModal(InsertSummaryOfChangesModal, {
+			submissionId,
+			reviewRounds,
+			currentLocale: getCurrentLocale(),
```
jardakotesovec: getCurrentLocale would give you currently selected UI locale. What we want to follow for submission/publication/file metadata is "primary locale for the submission". Which is available in submission.locale

cc @bozana  Hi Bozana, I think I got this right.. but tagging you to confirm. Thanks!

src/composables/useInsertSummaryOfChangesContent.js
```diff
@@ -0,0 +1,73 @@
...
+					([locale, html]) => {
+						if (!html) return;
+						const editor = editorsByLocale[locale];
+						if (!editor) return;
+						// Collapse selection to the end so insertContent appends.
+						const range = editor.dom.createRng();
+						range.selectNodeContents(editor.getBody());
+						range.collapse(false);
+						editor.selection.setRng(range);
+						editor.insertContent(html);
```
jardakotesovec: Could we use setValue? I think that would be bit simpler and cleaner than tracking the tinyMCE instances and interacting with all of them.

Something like:
```
			onInsert: (summaryOfChangesByLocale) => {
				const current = getValue(fieldName) ?? {};
				const next = {...current};
				Object.entries(summaryOfChangesByLocale || {}).forEach(
					([locale, html]) => {
						if (!html) return;
						next[locale] = (current[locale] || '') + html;
					},
				);
				setValue(fieldName, next);
				closeSideModal(InsertSummaryOfChangesModal);
			},
```

src/pages/workflow/workflowStore.js
```diff
@@ -31,6 +31,10 @@ export const useWorkflowStore = defineComponentStore(
 		const dashboardPage = props.pageInitConfig.dashboardPage;
 		const versionStageOptions =
 			props.pageInitConfig.componentForms.versionStageOptions || [];
+		const updateTypeOptions =
+			props.pageInitConfig.componentForms.updateTypeOptions || [];
```
jardakotesovec, line 35: Eventually we will need to figure out how to replace it with API, but thats out of scope.

jardakotesovec (changes requested):
Couple adjustments needed, truly important part is the locale selection.

src/composables/useInsertSummaryOfChangesContent.js
bozana: Hmmm... I am not sure what is this for... It seems it is only used to pick which locale's text to display as the preview in the modal list? What if there is no text in this locale? Because this field is not a required field, it can be that the entries exist for any locale that is not necessarily submission locale or current UI locale...
Shouldn't the localized function -- localizeSubmission(summaryByLocale, submissionLocale) -- be considered there instead? 
Or should it only reflect the one concrete, UI locale, even if there is no text entered in that locale?
I am only thinking loud because I do actually not know what this currentLocale mean... 
Hmmm...

bozana: Hmmm.... Now I see it in the UI...
"Insert Content" is added only to the TinyMCE editor for the journal primary locale field — not the UI locale, not the submission locale. Is this correct? Why this?
It then also inserts the same text in all publication's summaryOfChanges fields no matter which locale. Do we want it so?

bozana: Isn't it possible to have "Insert Content" in all TinyMCE fields and to list all existing summary of changes texts and the editor then chooses which to insert into which field?
Or to try to mach the locales the text exists in to the locales of the TinyMCE fields?
Also, the editor chooses one text -- that could be in the wrong language because we display only one language -- and we copy that text into all TinyMCE fields, for all locales. why do we want to copy one and the same text into all locales automatically? For usage later (e.g. display on the article landing page or downstream services) we can use localize() function to get the text in a specific locale, taking first UI locale, then submission locale, then other locales into account... 🤔 
So It seems to me that the firs option here -- to have "Insert Content" in all TinyMCE fields and to list all existing summary of changes texts and the editor then chooses which to insert into which field -- and not to copy anything automatically -- should work best... No?

src/composables/useInsertSummaryOfChangesContent.js
blesildaramirez: This tried to resolve the issue when clicking Insert and closes the modal, if the cursor previously was focused on the Multilingual buttons (French/English) it tries to set the focus there again.. so I wanted to ensure that it scrolls back to the Summary of Changes. But that didn't always work right, so I will remove this (as based from your suggestion above) and then just keep the current behavior since it's not always the case. We can try to address on a separate issue, as it's also happening when editing an email template.

src/composables/useInsertSummaryOfChangesContent.js
jardakotesovec: @bozana So when copying over things from the revised file amendment note to the publication amendment note.  Both fields are considered as metadata and follow the same set of locales, right? 

Intention from UX point of view to avoid inserting for each language individually, but copy things over for all languages at the same time, since both fields follow set of metadata locales and should match. Similarly like if we have reviewer recommendation and we want to pre-fill the reviewer details we also copy over all languages.

Per my test today it did not copy the same thing to all locales..  but it was copying individual locales to individual locales. That at least would be intended behaviour.

Also reason why there is just one locale passed to that insert modal is that we would be displaying just one locale there (submission primary locale). That was intentional decision.

If you are still concerned with this approach I can have one more round with Devika to see whether we can eliminate some of the concerns.   But would be handled as separate ticket so we don't hold this up.

bozana: Hi @jardakotesovec, oh yes, my bad -- somehow I have not seen that the locales are matched because I had the same text for all locales.
But: 
The button "Insert Content" is displayed only for the journal primary locale. Shouldn't it be the submission locale? 
Also, I think we should then use something like `localizeSubmission(summaryOfChanges)` to get the text that is then displayed in the modal: the text could be in any language not necessary submission locale -- the system does not require it to be in the submission locale. In reality it would probably be in the submission language, but I think we should consider the strange cases too i.e. how the system works? Thus, if journal primary locale, UI and submission locale are EN, but the summary of changes is entered only in FR, clicking on that button will not show anything. If the `localizeSubmission` behaves in the same way as `getLocalizedData()` behaves for submission object in PHP, it will first look if the data exists in the UI locale, then fall back to submission locale and then any other locale. So that the text is then always displayed. It will then be mapped and copied to the correct locale, I suppose.

bozana: I see Blessi has changed the button to appear only for submission locale... So this seems to be fine now... Sorry!

bozana: Then only one question remains for me: the text is now displayed in the submission locale only. Do we not want to display it if it exists in any other locale but not in the submission locale?

PR #888, opened 2026-05-13
pkp-pkp-lib#12262 Improve select field and simplify table rowspan rendering

blesildaramirez:


src/components/SelectInput/SelectInput.vue
```diff
@@ -27,26 +20,26 @@
 			</option>
 		</select>
 		<Icon
+			v-if="isBorderless"
```
jardakotesovec, line 23: This is probably also good solution. I have not attempted to customise significantly look of the select element, so not sure how much fighting potentially it is. Don't have strong preference either way - but its also possibility to use the Select component from reka-ui. I guess that becomes beneficial especially if there is lots of visual customisation.  I see that in new design that this icon is used consistently in new forms, so eventually this could be simplified here.

src/components/Table/TableRow.vue
```diff
@@ -3,42 +3,34 @@
...
 const groupId = inject('groupId', null);
-const rowSpanGroupContext = inject('rowSpanGroupContext', null);
+const rowSpanGroupSize = inject('rowSpanGroupSize', null);
 
-defineProps({
+const props = defineProps({
 	/** Enables striped styling for the row */
 	striped: {type: Boolean, default: true},
+	/** Marks this row as a continuation of a row-spanning group (not the first row). Cells with rowspan > 1 will be skipped. */
+	isContinuation: {type: Boolean, default: false},
```
jardakotesovec, line 24: What was motivation to move the isContinuation to the props api?

jardakotesovec (changes requested):
Feels that exposing isContinuation as props and being controlled from outside is not very intuitive.. but interested to know your take as I am no in that level of detail.

src/components/Table/TableRow.vue
blesildaramirez: Hi Jarda, the reason for this is to avoid complicating the use of register/unregister. The parent knows when to use the `rowspan`, but I agree handling it outside is not intuitive. I can revert back to it and remove this as prop, and have a simpler approach on detecting when to render the cell if it belongs to a rowspan.

blesildaramirez:
Closing, improvements addressed on this new PR: https://github.com/pkp/ui-library/pull/903

PR #889, opened 2026-05-14
pkp/pkp-lib#11825 Apply template variables api to use for insert content in task templates

blesildaramirez:


jardakotesovec (approved):
Looking good :-).

PR #894, opened 2026-05-28
Review multiselect

jardakotesovec:


blesildaramirez (approved):
Everything looks good on my end. Also acknowledging that that this new field is only added in OJS, and not OMP & OPS.

ewhanson:
Merged as part of pkp/ui-library#951.

PR #904, opened 2026-06-10
pkp/pkp-lib#12799 Support submission removal from workflow queues

ewhanson:


src/pages/workflow/composables/useWorkflowDecisions.js
```diff
@@ -23,6 +23,8 @@ export const Actions = {
 	DECISION_BACK_FROM_PRODUCTION: 'decisionBackFromProduction',
 	DECISION_RECOMMEND_ACCEPT: 'decisionRecommendAccept',
 	DECISION_RECOMMEND_DECLINE: 'decisionRecommendDecline',
+	DECISION_MOVE_TO_DONE: 'decisionMoveToDone',
```
bozana: Maybe we do not need this?

Claude says:
```
 export function wrapActionFns(Actions, ActionsFns, wrapperFn) {
      return Object.values(Actions).reduce((acc, actionName) => {
          acc[actionName] = (actionArgs, finishedCallback) =>
              wrapperFn(ActionsFns[actionName], actionArgs, finishedCallback);
          return acc;
      }, {});
  }

Object.values(Actions) takes every value from the Actions object, so for DecisionActions that includes 'decisionMoveToDone'. For each value it creates an entry in the result object that calls ActionsFns[actionName] — i.e. ActionsFns['decisionMoveToDone'] — which is undefined since no such function exists.

  And this is called in workflowStore.js line 181:

  const decisionActions = wrapActionFns(
      DecisionActions,       // ← includes DECISION_MOVE_TO_DONE
      _workflowDecisionsFns, // ← does NOT have decisionMoveToDone
      ...
  );
```

PR #905, opened 2026-06-16
I12897

jardakotesovec:


src/components/PandocConverter/pandocLoader.js
```diff
@@ -0,0 +1,33 @@
...
+import {createPandocInstance} from '../../../node_modules/pandoc-wasm/src/core.js';
+import pandocWasmUrl from '../../../node_modules/pandoc-wasm/src/pandoc.wasm?url';
+
+let pandocPromise = null;
+
+export function loadPandoc() {
+	if (pandocPromise) return pandocPromise;
+	pandocPromise = fetch(pandocWasmUrl)
+		.then((r) => r.arrayBuffer())
+		.then((buf) => createPandocInstance(buf));
```
blesildaramirez: To be safe, we can add here a `catch` to ensure errors are handled, also to reset `pandocPromise` in case of an error:

```suggestion
		.then((buf) => createPandocInstance(buf))
		.catch((err) => {
			pandocPromise = null;
			throw err;
		});
```

blesildaramirez (approved):
I just have one minor suggestion for error handling, after addressing it please feel free to merge. Thanks!

src/components/PandocConverter/pandocLoader.js
```diff
@@ -0,0 +1,33 @@
...
+ * So we reach the package's internal core.js via a direct relative
+ * node_modules path. core.js exposes `createPandocInstance` (a plain factory,
+ * no top-level await), and we fetch the .wasm ourselves via Vite's `?url`.
+ * This also loads the ~58 MB binary lazily — only when an import actually runs,
+ * not on every page load (which the official entry would do).
+ *
+ * Caveat: core.js is an internal module, not part of pandoc-wasm's public API,
+ * so re-verify these imports when upgrading the dependency.
+ */
+import {createPandocInstance} from '../../../node_modules/pandoc-wasm/src/core.js';
```
blesildaramirez: I understand the need to import this directly through the `node_modules` directory given the bundling/exports constraints. Also noted on re-verifying when upgrading the dependency.

Maybe we can put a comment here to include the current `pandoc-wasm` version as well, so it matches what's in `package.json`, and makes it easier to remember to recheck this when we upgrade.

src/components/PandocConverter/pandocLoader.js
```diff
@@ -0,0 +1,33 @@
...
+import pandocWasmUrl from '../../../node_modules/pandoc-wasm/src/pandoc.wasm?url';
```
blesildaramirez: Maybe we can add an `alias` to our `vite.config.js` file: 

```
    alias: { 
         'pandoc-core': path.resolve(__dirname, 'node_modules/pandoc-wasm/src')
     } 
```

Then these imports can be updated to

```suggestion
import {createPandocInstance} from 'pandoc-core/core.js';
import pandocWasmUrl from 'pandoc-core/pandoc.wasm?url';
```

blesildaramirez (approved):
Adding a couple more comments to this PR. Up to you if you would apply these suggestions..

PR #907, opened 2026-06-17
pkp/pkp-lib#7527 stamp identity metadata

bozana:
s. https://github.com/pkp/pkp-lib/issues/7527

### DIFF
--- a/src/pages/workflow/WorkflowPage.vue
+++ b/src/pages/workflow/WorkflowPage.vue
@@ -29,6 +29,26 @@
 					)
 				"
 			></span>
+			<div v-if="selectedPublication.contextName" class="text-lg-normal">
+				<span
+					v-strip-unsafe-html="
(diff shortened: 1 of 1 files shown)


kaitlinnewson (approved):

src/pages/workflow/WorkflowPage.vue
```diff
@@ -29,6 +29,26 @@
 					)
 				"
 			></span>
+			<div v-if="selectedPublication.contextName" class="text-lg-normal">
```
kaitlinnewson, line 32: Do we need to display this at the top of the workflow page, given that it is already displayed in the Publication Settings tab and may be the same as the defaults? It feels crowded with that amount of information and may need a bit more UI design work (e.g. the pipe characters need some accessibility tweaks and the vertical spacing could be increased).

Image

PR #918, opened 2026-06-23
pkp/pkp-lib#12222 Include review types data in participant options for reviewer roles

blesildaramirez:


src/components/Form/display/FieldOptionsDisplay.vue
```diff
@@ -12,9 +12,7 @@
 				<template v-if="field.showNumberedList">{{ i + 1 }}.</template>
 				{{ option.label }}
 			</span>
-			<span v-if="option.subLabel">
-				{{ option.subLabel }}
-			</span>
+			<span v-if="option.subLabel" v-strip-unsafe-html="option.subLabel" />
```
jardakotesovec: What about to add subLabel2. Not ideal. But I see that we currently don't have option to define custom display component for given field..  so maybe this could be slightly better than injecting html?

jardakotesovec (changes requested):

src/components/Form/display/FieldOptionsDisplay.vue
blesildaramirez: Hi @jardakotesovec , this is now updated. Thanks!

jardakotesovec (approved):

PR #925, opened 2026-07-09
pkp/pkp-lib#13011 Update citation manager to use nested submission/pu…

bozana:
…blication citation API routes

see https://github.com/pkp/pkp-lib/issues/13011


touhidurabir (approved):
@bozana looks good to me .

PR #937, opened 2026-07-22
pkp/pkp-lib#12932 Trigger search when pressing Enter key

blesildaramirez:
- This updates all the tables to make the behavior of the search consistent. The search should trigger only when the Enter key is pressed.
- Clear Filters will only clear all the filters, leaving the search text untouched.
- Minor fix to show the x clear search button in Structured References table.


jardakotesovec (approved):
Have not tested, but nothing suspicious here :-)

PR #938, opened 2026-07-23
pkp/dev-team#285 [ORE Submission] Move the location of the Data Availability block in the form

blesildaramirez:
Notes:
- Submission wizard has multiple panels which could have multiple forms in it. These panels don't have the group logic since they are not in a single form, so I added a mapper in the `SubmissionWizardPage` where vue logic are handled.
- The `-m-5` class in WorkflowPublicationForm.vue caused rendering issues on the Data Availability form. The section shows directly after the table (without padding) because of this. Easiest fix is to pass class "mb-5" as prop in Data Citation table in the workflow config. I tried several approaches, like passing prop instead to remove the `-m-5` conditionally, but this caused another issue which added a big padding in the form for Data Availability. So I'm left with this one for now, we can go probably back later if this needs to be improved.

### DIFF
--- a/src/components/Container/SubmissionWizardPage.vue
+++ b/src/components/Container/SubmissionWizardPage.vue
@@ -125,9 +125,7 @@ export default {
 		/**
 		 * A step's sections grouped for display, keyed by step id
 		 *
-		 * Sections sharing a `group` key render in one panel, using the first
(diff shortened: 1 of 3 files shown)


blesildaramirez:
Merged via https://github.com/pkp/ui-library/pull/939

PR #950, opened 2026-07-29
pkp/pkp-lib#13094 [Reviewer's Page] Make submission details available to reviewers

blesildaramirez:
Note of the changes:
- Added 3 form display components (field-authors, field-controlled-vocab, field-html)
- Added a new View modal for the citations table


src/pages/reviewerSubmission/ReviewerSubmissionDetailsModal.vue
```diff
@@ -0,0 +1,144 @@
...
+	disciplines: {type: Array, default: () => []},
+
+	/** Not visible to double-anonymous reviewers */
+	authors: {type: String, default: null},
+	dataAvailability: {type: String, default: null},
+	fundingStatement: {type: String, default: null},
+	dataCitations: {type: Array, default: () => []},
+
+	/** Legacy options automatically added by modalStore when opened from a legacy modal handler */
+	legacyOptions: {type: Object, default: () => ({})},
```
jardakotesovec: Is this one needed?

src/pages/reviewerSubmission/ReviewerSubmissionDetailsModal.vue
```diff
@@ -0,0 +1,144 @@
...
+				</div>
+
+				<div v-if="abstract">
+					<h2 class="mb-1 text-xl-bold text-default">
+						{{ t('common.abstract') }}
+					</h2>
+					<div v-strip-unsafe-html="abstract" class="text-lg-normal"></div>
+				</div>
+
+				<PkpTable v-if="dataCitations.length">
```
jardakotesovec: Could we reuse the data citation manager? 

If I am not mistaken we discussed that there would be Read action which would show the data citation detail in the modal - same form as currently its there.. just using the 'display only' mode?

src/pages/reviewerSubmission/ReviewerSubmissionDetailsModal.vue
```diff
@@ -0,0 +1,144 @@
...
+					heading-element="h2"
+					:heading="t('submission.fundingStatement')"
+					:html-value="fundingStatement"
+				></FormDisplayItemBasic>
+
+				<PkpTable
+					v-if="metadataRows.length"
+					:aria-label="t('submission.informationCenter.metadata')"
+				>
+					<TableHeader>
```
jardakotesovec: Can't we just use form display for this? Not sure where the table layout came from? 

There might have been some confusion around the designs?

src/pages/reviewerSubmission/ReviewerSubmissionDetailsModal.vue
blesildaramirez: Not really, I removed it. It's something our ajax modals send which is useful for handling events, which is not needed in the view submissions modal since it's readonly. Thanks for pointing it out.

blesildaramirez:
Hi @jardakotesovec , suggestions have been applied to this PR. Thanks!

src/pages/reviewerSubmission/useReviewerSubmissionDetailsForm.js
```diff
@@ -0,0 +1,120 @@
...
+
+		const fundingStatement = localize(publication.fundingStatement);
+		if (fundingStatement) {
+			addFieldHtml('fundingStatement', {
+				label: t('submission.fundingStatement'),
+				value: fundingStatement,
+			});
+		}
+
+		const keywords = localize(publication.keywords);
```
jardakotesovec, line 91: I actually think that showing it multilingual would be pretty cool..  or at least being able to pass the multilingual data and select the language.

But probably needs more thinking what the behaviour exactly should be..  Ideally it would detect the languages which are actually used in the fields.. and than it would present it in these languages..  But this would need bit more thinking..  so I think for now probably best to keep it as is..

src/managers/DataCitationManager/useDataCitationManagerConfig.js
```diff
@@ -12,12 +12,15 @@ export function useDataCitationManagerConfig() {
...
 		});
 
-		columns.push({
-			header: t('common.moreActions'),
-			headerSrOnly: true,
-			component: 'DataCitationManagerCellActions',
-			props: {},
-		});
+		// Skip the actions column when view-only
+		if (canEdit?.value) {
```
jardakotesovec: I was hoping that we would have read only content in the modal as well - we would have 'View' action instead of 'Edit'.  But maybe there are some challenges with that as there is couple custom components..  so it would not be just switching to read only. So depending on the feasibility - it would be useful if they can get to individual informations there..

blesildaramirez:
Hi @jardakotesovec , this PR is ready for code review based on your latest feedback. Could you take a look? Thanks!

src/managers/DataCitationManager/useDataCitationManagerActions.js
```diff
@@ -36,6 +38,26 @@ export function useDataCitationManagerActions() {
...
 
+	function dataCitationViewDataCitation({dataCitationEditForm, dataCitation}) {
+		if (!dataCitation.authors) {
+			dataCitation.authors = [];
+		}
+
+		const viewForm = cloneDeep(dataCitationEditForm);
+
+		// Display mode still renders the footer, so drop the buttons it is built from
+		viewForm.pages?.forEach((page) => {
```
jardakotesovec, line 49: Would not be better just don't render the buttons in display only mode? Or do we already have some case where we need buttons even with the display only mode?

src/managers/DataCitationManager/useDataCitationManagerConfig.js
```diff
@@ -44,17 +44,25 @@ export function useDataCitationManagerConfig() {
 		const actions = [];
 
 		actions.push({
-			label: t('common.edit'),
-			name: Actions.DATA_CITATION_EDIT_DATA_CITATION,
-			icon: 'Edit',
+			label: t('common.view'),
```
jardakotesovec, line 47: Thats interesting.. did not think of that. So we would have the 'View' action always available. I would be curious what Devika thinks - to me that seems reasonable.

src/components/Form/display/FieldAuthorsDisplay.vue
```diff
@@ -0,0 +1,53 @@
...
+				<span v-if="index > 0">{{ t('common.commaListSeparator') }}</span>
+				<span class="inline-flex items-center">
+					<span v-if="author.name">{{ author.name }}</span>
+					<a
+						v-if="author.orcid"
+						class="inline-flex items-center"
+						:href="author.orcid"
+						target="_blank"
+					>
+						<Icon icon="Orcid" class="ms-1 w-4" aria-hidden="true" />
```
jardakotesovec: Since these orcid are not authenticated - probaby best to use the `OrcidUnauthenticated` icon

jardakotesovec (approved):
Just couple minor questions/suggestions, but overall good to go.

src/managers/DataCitationManager/useDataCitationManagerActions.js
blesildaramirez: Hi Jarda, yes there's a case in tasks & discussions that we render footer buttons in display mode.

src/managers/DataCitationManager/useDataCitationManagerConfig.js
blesildaramirez: Good point, I will ask Devika about this UI addition

PR #960, opened 2026-08-20
pkp/pkp-lib#13156 [Editorial UI] - Allow journal editors to modify reviews

blesildaramirez:
Changes:
- New components to show "Review Details" & "Modify Review" form
- New display component: "FieldRichTextareaDisplay"
- Added support for Reviewer's uploaded fles in File Manager
- Reordered list of action buttons in the Reviewers table


src/managers/FileManager/useFileManagerConfig.js
```diff
@@ -367,6 +367,50 @@ export const FileManagerConfigurations = {
...
+			Actions.FILE_UPLOAD,
+			Actions.FILE_EDIT,
+			Actions.FILE_DELETE,
+			Actions.FILE_SEE_NOTES,
+		],
+		fileStage: pkp.const.SUBMISSION_FILE_REVIEW_ATTACHMENT,
+		titleKey: tk('reviewer.submission.reviewerFiles'),
+		wizardTitleKey: tk('common.upload'),
+	}),
+	REVIEWER_ATTACHMENT_FILES_READ_ONLY: () => ({
```
jardakotesovec: For this use case - I am not sure whether new configuration is good fit. As this 'read only' might be useful for another places where we just replay the choices. 

What about having separate readOnly flag - that would just filter out the editing actions when the permissions are calculated?

src/managers/ReviewerManager/useReviewForm.js
```diff
@@ -0,0 +1,105 @@
...
+			const {component, name, type, options, ...fieldProps} = field;
+
+			const commonFields = {
+				...fieldProps,
+				groupId,
+				isRequired: inDisplayMode ? false : !!field.isRequired,
+				value: getResponseValue(field, reviewFormResponses),
+			};
+
+			switch (component) {
```
jardakotesovec: Why do we need this mapping? I think api is already sending ready to use form config?

src/managers/ReviewerManager/ReviewDetailsRating.vue
```diff
@@ -0,0 +1,110 @@
...
+const descriptionId = useId();
+
+// Undefined until the review loads, when null means the editor left it unrated
+function toSelection(rating) {
+	return rating === undefined ? null : (rating ?? 0);
+}
+
+const quality = ref(toSelection(props.reviewAssignment.quality));
+
+watch(
```
jardakotesovec, line 69: This is not bad solution - only downside is that client trust the server - if it says ok, it trust it saved the rating correctly. Which is already pretty good strategy.  But what we are aiming when possible is to reload the data after updating - and that use as source of the truth. In this trivial example it does not make that much difference - but in general benefit is to have just one path how to get data from server & and just display them - without carrying additional logic how to update the data client side, which is always bit of the duplication and risk for getting out of the sync with server. Hope its clear what I have in mind.

So I am wondering whether the logic for rating update could go to the parent component so its easy to reload the review assignment after the update and keep this component simpler - responsible just for rendering and when editor change the rating it would emit that change for parent to handle.

src/managers/ReviewerManager/useReviewDetailsForm.js
```diff
@@ -0,0 +1,453 @@
...
+			reviewContentRef.value = reviewContent;
+			reviewComments.value = reviewContent?.comments ?? '';
+			addReviewContentFields(reviewContent);
+		}
+
+		setInitialState();
+	}
+
+	// Get review comments/form if set
+	async function loadReviewContent() {
```
jardakotesovec: This one is pretty complex, bit difficult to follow the contract between useReviewDetails and useReviewDetailsForm 

I am not sure whats the best way to organise this. Usually what helps is to have top layer focused on orchestration and fetching data and the lower layers just on displaying and reporting the changes to the parent. 

I see for example reviewAssignment being passed to useReviewDetailsForm - which is than exposing it again as reviewAssignmentRef? 

If you think that this organisation works well, feel free to push.. it might be just me not following it well..

jardakotesovec (changes requested):
First round of higher level suggestions to think about.  I expect to do another round once its the APIs are in place.. and ideally when its bit easier to follow.. to properly look into the business logic details..

src/managers/FileManager/useFileManagerConfig.js
blesildaramirez: Hi Jarda, updated this to use `readOnly` prop. Thanks for the suggestion.

src/managers/ReviewerManager/useReviewForm.js
blesildaramirez: Because we didn't have an `addField` in our `useForm` composable. We have `addFieldText`, `addFieldSelect`, etc - hence the mapping. I added the `addField` instead so the whole field config is passed, along with the field name.

src/managers/ReviewerManager/ReviewDetailsRating.vue
blesildaramirez: Good point - I have updated this to always refresh when `reviewAssignment` changes, which is triggered to reload after a successful `PUT` on the endpoint.

src/managers/ReviewerManager/useReviewDetailsForm.js
blesildaramirez: I agree - sorry about that. I've reorganized so each API endpoint has its own composable: useReviewAssignment for everything on `/reviewAssignments/{id}` (the GET, the rating PUT, and confirm/mark-viewed on /consider), and `useReviewContent` for `/review` - loading it, saving it, and the downloads.
Here's the mapping for each imported apis:
- ReviewDetailsModal.vue -> useReviewDetails -> `useReviewAssignment, useReviewContent, useReviewDetailsForm`
- ReviewDetailsEditModal.vue -> useReviewDetailsEdit -> `useReviewContent, useReviewDetailsForm`
- useReviewDetailsForm -> useReviewContentFields

`useReviewContentFields` is split out so the reviewer's Step 3 form can reuse it later.

src/managers/ReviewerManager/useReviewAssignment.js
```diff
@@ -0,0 +1,135 @@
...
+
+		await sendRating();
+		const isSaved = isRatingSaved.value;
+
+		await loadReviewAssignment();
+
+		return isSaved;
+	}
+
+	return {
```
jardakotesovec, line 124: that looks good!

jardakotesovec: I was wondering for moment what isConfirming means.. but I think I get it :-)..

src/managers/ReviewerManager/useReviewDetailsEdit.js
```diff
@@ -0,0 +1,116 @@
...
+	recommendations = [],
+	onSavedFn = () => {},
+}) {
+	const {t} = useLocalize();
+	const {openDialog} = useModal();
+	// Modals mount outside the workflow page, so the file manager needs its own provider
+	useDataChangedProvider();
+	const closeModal = inject('closeModal');
+
+	const reviewAssignmentRef = ref(reviewAssignment);
```
jardakotesovec, line 29: I wonder whether this work as intended..   as I suspect when you pass props to the useReviewDetailsEdit and than destruct it to reviewAssignment - at that moment reviewAssignment is not reactive. 

I think you would need to do something like:
```javascript
export function useReviewDetailsEdit(props) {
  const {
    submission,
    submissionStageId,
    reviewRoundId,
    reviewAssignment,
  } = toRefs(props)
```

src/managers/ReviewerManager/useReviewDetailsForm.js
```diff
@@ -0,0 +1,227 @@
...
+			submissionStageId,
+			// Unused for filtering, but the legacy upload wizard requires it
+			reviewRoundId,
+			reviewAssignmentId: reviewAssignment.value.id,
+		},
+		groupId: 'reviewerFiles',
+	});
+
+	if (isOJS()) {
+		addGroup('reviewRecommendation', {
```
jardakotesovec, line 177: I though that OMP has reviewRecommendation - but its not required-  can you double check?

src/managers/ReviewerManager/useReviewContentFields.js
```diff
@@ -0,0 +1,162 @@
...
+ * Builds a submitted review into form fields: a review form's questions, or textarea comments.
+ */
+export function useReviewContentFields(
+	{addGroup, addField, addFieldRichTextArea, addFieldComponent},
+	{inDisplayMode = false, canEditPrivateComment = false} = {},
+) {
+	const {t} = useLocalize();
+
+	/** Each field type needs its own empty value, or it renders as answered */
+	function getResponseValue(field, reviewFormResponses) {
```
jardakotesovec: As discussed on mattermost - lets have a look whether the normalisation for setting values in useForm could be leveraged, so we would not need custom mapping.

jardakotesovec (commented):
Thanks for the updates, looks great.

Just couple small things to check.

src/managers/ReviewerManager/useReviewContentFields.js
blesildaramirez: For using `setValues`, it only writes the field names present in the object you pass it. The `/review` API always sends `value: null`, and that's what goes into `addField` initially - so if we called `setValues` later with the `reviewFormResponses`, only the answered questions would get a value and the rest would stay `null`. 
`null` isn't a valid empty value for every field type: a checkbox group needs `[]`, a select needs `null`, a textarea needs `''`. So we'd still need something like `getResponseValue` to fill those in.

`setValues` only sets the fields you pass to it, so we'd want it to also fill in the unanswered fields with an empty value. It doesn't do that today because `useForm` determines the empty value based on what the field currently holds (`getClearValue`). In this case, that's `null` for all the review form fields, so a checkbox group would end up as `''` instead of `[]`.
We could change it to determine the empty value based on the field type, but that would be a change to a shared composable and would need to be backwards compatible since there are nine callers on the `setValues` that only touch the keys they pass.
I think that's worth doing later, but probably not in this branch. It also wouldn't handle the checkbox conversion below.

On the checkboxes: reviewer answers submitted through the Step 3 form are stored as an array of strings, and the API returns them as-is. Answers saved through the new API come back as integers. Since the option values are numeric, we need the mapper so both cases match.

src/managers/ReviewerManager/useReviewDetailsForm.js
blesildaramirez: No, OMP doesn't render the recommendation in step 3.

Internal:
image

External:
image

src/managers/ReviewerManager/useReviewAssignment.js
blesildaramirez: Yes, `isConfirming` is for waiting until the `/consider` api is resolved.

src/managers/ReviewerManager/useReviewContentFields.js
jardakotesovec: Alright.. the root cause is server side not setting correct empty values..  even though FormComponent has concept of empty value - would this help? 
https://github.com/pkp/pkp-lib/pull/13233/changes/57ed6b1031a922cd9ac470d1bcf870596a2f58be

src/managers/ReviewerManager/useReviewDetailsForm.js
jardakotesovec: I see, thank you for confirming

src/managers/ReviewerManager/useReviewDetailsForm.js
jardakotesovec: Great!

src/managers/ReviewerManager/useReviewContentFields.js
jardakotesovec: Also tagging @taslangraham on his opinion..

jardakotesovec: Maybe it could be addressed directly in FormComponent where is the emptyValue logic - but currently its probably not efficient..  as the if (!array_key_exists('value', $config))  is not doing anything - since the value is always set to something in the config already...   Which is probably unintended..  

So maybe best to try whether using getEmptyValue helps.. and than think about where that should be applied ideally..

src/managers/ReviewerManager/useReviewDetailsEdit.js
blesildaramirez: They behave the same here. The Modify modal gets `reviewAssignmentRef.value`, which is already a resolved object, and the prop doesn't change while the modal is open. So `toRefs` wouldn't really give us anything here.

The ref isn't needed to track changes to the prop. `useReviewDetailsForm` is also used by the View modal, where the assignment can change as it loads, so it expects `reviewAssignment.value`. Here, we just need to give it something with a `.value`.

I'm happy to switch to `toRefs` if you prefer, but it wouldn't change anything in this case.

src/managers/ReviewerManager/useReviewContentFields.js
blesildaramirez: I checked, and most of the `getResponseValue` will go away with https://github.com/pkp/pkp-lib/commit/57ed6b1031a922cd9ac470d1bcf870596a2f58be - thanks for that.
What it doesn't cover is the checkbox inconsistency. Both save paths go through `saveReviewFormResponse` and store the value as submitted, but the Step 3 form posts strings while the API sends numbers, so the same field can come back as `["1","2"]` or `[1,2]`. Since the option values are numeric, we still need to normalise that.

So the mapper shrinks to just the checkbox case, and with the empty values now coming from the API I can switch to `setValues` as you suggested once we merge the commit you provided above.

src/managers/ReviewerManager/useReviewDetailsEdit.js
jardakotesovec: Oh I see.. so it was not to actually have reactivity in this case - but to satisfy that the '.value' is expected..  In that case nevermind.. maybe small comment as it looks like intention to keep the reactivity in case reviewAssignment changes

src/managers/ReviewerManager/useReviewContentFields.js
jardakotesovec: I need to run now.. so can possibly look tomorrow if you or @taslangraham don't beat me to it. But inconsistency with the "1" vs 1 - that also deserves some checking where that inconsistency happens..  to make sure that its always the same..

taslangraham: @jardakotesovec, your fix for empty values makes sense to me. I think the `saveReviewFormResponse` method should normalize submitted checkbox values to numbers (`["1","2"]` gets stored as `[1,2]`). It might be worth normalizing values for radio button and select options too. The new endpoint returning review form information should also ensure values are normalized to numbers; this ensures that already existing string values for checkboxes submitted by the Step 3 form will be exposed as numbers, thus keeping everything exposed via the `/review` endpoints consistent.

taslangraham: @blesildaramirez I think this should suffice: https://github.com/taslangraham/pkp-lib/commit/45c4a0fc7c10cc4b8e071ad0cd20ff894bfd519a for the normalization issue.

blesildaramirez: Hi @taslangraham , thanks! The `Repository.php` changes and the read/write split look right to me.

One thing with `getReviewFormResponses`: the loop only handles `is_array($value)`, so text, textarea, radio, and dropdown responses get dropped. They would come back missing and show as unanswered in the modal. Could you verify this one?

With those that fixed, `getResponseValue` can go away completely on my side rather than just being reduced.

taslangraham: @blesildaramirez, it looks like I accidentally omitted those responses in my fix. I've updated the code.

blesildaramirez: Hi @taslangraham , thanks ! The changes look good to me, I tested and yours and Jarda's commit are working correctly on my end. I think yours can be merged to pkp-lib. 
Also hi @jardakotesovec , with both commits I can remove the `getResponseValue` mapping on my end, please feel free to also merge your commit.

Thanks both!

taslangraham: Hi @blesildaramirez, my fix is now available on pkp-lib `main` branch. You can rebase your pkp-lib local branch to get the updates.

jardakotesovec: Thank you for tackling int/string  - my understanding is that field choices are always recorded as integers - so the checkboxes being strings in array were some old inconsistency. So I think that was good call to fix - so its now always integers.

This change also means that now we have both versions in the existing databases.  So feels like making it consistent via migration would be nice clean up. Currently did not find any broken consumers.. but could surprise someone processing these data.

@asmecher 


 Its relatively minor inconsistency, but I think worthwhile to address - what you think? On the screenshot you can see it before and after.

Image

jardakotesovec: Here is my attempt to properly address the empty values not being set..   intend to merge it once the tests pass
https://github.com/pkp/pkp-lib/issues/13237

src/managers/ReviewerManager/useReviewDetailsEdit.js
blesildaramirez: Hi Jarda, I added a comment to the `ref` usage.

PR #971, opened 2026-09-01
Set recommendation options when opened from submissions listing

blesildaramirez:


Tribunal33:
~~/deploy~~ forgot it's a submodule.  N/M

PR #972, opened 2026-09-01
pkp/pkp-lib#13242 Update frontend components to support Blade theme

NateWr:


src/frontend/components/PkpComments/docs/PkpComments.example.css
```diff
@@ -26,16 +26,6 @@
 	align-items: center;
 }
 
-.PkpCommentsNotificationNotLatest {
```
NateWr, line 29: I started some work on adapting the styles, but didn't finish. The storybook probably needs a fair amount of updating to match the changes to comments and open peer review.

src/frontend/components/PkpOpenReviews/docs/PkpOpenReview.stories.js
```diff
@@ -1,5 +1,5 @@
 import '@/styles/frontend-theme.css';
-import PkpOpenReview from '../PkpOpenReview.vue';
```
NateWr, line 2: Although I no longer use the Vue component to display open reviews, I left all of my changes in, in case you want to use the Vue component to in ORE.

If not, you can remove all of the related code.

src/frontend/components/PkpOpenReviews/icons/Approved.vue
```diff
@@ -0,0 +1,14 @@
+<template>
```
NateWr, line 1: All of these icons are now included in the Eidos blade templates instead. I left them here for the component, but feel free to remove them once you are ready.

src/frontend/components/PkpOpenReviews/docs/PkpOpenReview.stories.js
jardakotesovec: I will be interested to look into blade version..  to see how feasible is to bring ore there. But don't understand why you would be refactoring vue.js version at all? Its already been adapted to align with eidos design?

PR #976, opened 2026-09-02
pkp/pkp-lib#13109 publication-based check if user can edit publication

Vitaliy-1:


blesildaramirez (approved):
Looks good, approved.

PR #985, opened 2026-09-14
Remove unnamed file manager namespaces from workflow stage options

blesildaramirez:
The "Done" stage now appears in the submission's list of workflow stages. When attaching files from a workflow stage in a task or discussion, it showed up as a disabled option in the stage dropdown.

The stages come from the stages property returned by the /submissions/{id} API, which now includes "Done". This stage has no file manager configuration, so this PR uses that to filter the stages shown in the dropdown.

Another option would be a new property on each stage, e.g. `isWorkflowStage`, that says whether the stage is part of the workflow users see. But I think the property can be added if more places need it later.


src/components/FileAttacher/useFileAttacherWorkflowStage.js
```diff
@@ -13,13 +13,15 @@ export function useFileAttacherWorkflowStage(props) {
...
 	});
 
-	const options = props.submission?.stages.map((stage) => {
-		return {
-			label: stage.label,
-			value: stage.id,
-			disabled: !fileManagerUploadNamespaces[stage.id]?.length,
-		};
-	});
+	const options = props.submission?.stages
```
jardakotesovec: This looks good, just thinking whether we could make this more universally available for other possible use cases. What about introducing some helper in useSubmission.. something like ` getWorkflowStages(submission)` that would do the filtering?

blesildaramirez: Hi @jardakotesovec , thanks! I applied this suggestion, could you check again? Thank you!

jardakotesovec (approved):
Looks good.. I guess we just need to make sure that the WORKFLOW_STAGE_ID_DONE is exposed right from the backend..? I guess there is another PR for it :-)

blesildaramirez:
> Looks good.. I guess we just need to make sure that the WORKFLOW_STAGE_ID_DONE is exposed right from the backend..? I guess there is another PR for it :-)

Yes, WORKFLOW_STAGE_ID_DONE is already exposed in `PKPTemplateManager::setupBackendPage()` (`classes/template/PKPTemplateManager.php` in pkp-lib), so it's safe to use the constant in the frontend `getWorkflowStages()`

Merging this.

PR #988, opened 2026-09-17
pkp/pkp-lib#13353 display crossref registration message as html and remove registeredMessagePreamble

bozana:
see https://github.com/pkp/pkp-lib/issues/13353


src/components/ListPanel/doi/DoiItemViewRegisteredMessageDialogBody.vue
```diff
@@ -1,15 +1,13 @@
 <template>
 	<DialogBody>
-		<p>{{ registeredMessageAgencyPreamble }}</p>
```
ewhanson, line 3: Did you check how this works with Datacite and any other registration agency plugins? I don't know if it will make a big difference, but the assumption here was that this would be applicable across different types of plugins, but maybe that's something you've already considered and determined not to be a problem.

src/components/ListPanel/doi/DoiItemViewRegisteredMessageDialogBody.vue
```diff
@@ -1,15 +1,13 @@
...
-		<p>{{ registeredMessage }}</p>
+		<p v-strip-unsafe-html="registeredMessage" class="whitespace-pre-line"></p>
```
ewhanson, line 3: Another question about the preamble: the preamble text was added to make it clear that the text that followed was intended and the results that came back (when there was an error) rather than an error with rendering the webpage. With pre-3.4 Crossref errors, users often thought the XML that came back was not significant semantic content (e.g. the explanation of the error) but OJS encountering a rendering error since the XML error looked like HTML. Do you think this is clear enough now or would it be better to keep something like this?

ewhanson (changes requested):

src/components/ListPanel/doi/DoiItemViewRegisteredMessageDialogBody.vue
bozana: This changes only the success ("View record") dialog. The error dialog and its preamble are exactly as before. DataCite and mEDRA do not have this kind of message, they do not use this "View record".

bozana: A third-party agency -- if one would exist -- would just lose the preamble line and have its message rendered through `v-strip-unsafe-html`, which is ok for text/HTML messages.

src/components/ListPanel/doi/DoiItemViewRegisteredMessageDialogBody.vue
bozana: This changes only the success ("View record") dialog. The error dialog and its preamble are exactly as before, see `DoiItemViewErrorDialogBody.vue`. Thus, the error dialog is unchanged and still shows "The following error was returned by {agency}…" above the raw response in a `<pre>` block.

For the success dialog, the preamble was actually misleading. It said "The following message was returned by Crossref…", but what we store isn't something Crossref sent back. It's our own sentence (plugins.generic.crossref.successMessage):
```
msgid "plugins.generic.crossref.successMessage"
msgstr ""
"The metadata for this item has been deposited with Crossref. "
"To view further details, see the submission in the <a href=\"https://doi.crossref.org/servlet/submissionAdmin?sf=detail&submissionID={$submissionId}\">Crossref admin panel</a>."
```
With this change, when there are warnings we add another sentence of our own, not the XML.
Rendering it as HTML turns the admin-panel link into a real link instead of showing escaped `<a href=…>` text.

